PR #3, opened 2012-10-29
get_highest_res() method addition

lextoumbourou:
Submitting a proposal for a new method called get_highest_res(). It allows the user to automatically get the highest Youtube resolution. They can also specify the preferred file format, though it will favour highest quality over preferred format.

Here it is in action against your Readme example:

```
>>> from pytube import YouTube
>>> from pprint import pprint
>>> yt = YouTube()
>>> yt.url = "http://www.youtube.com/watch?v=Ik-RsDGPI5Y"
>>> pprint(yt.videos)
[<Video: MPEG-4 Visual (.3gp) - 144p>,
 <Video: MPEG-4 Visual (.3gp) - 240p>,
 <Video: Sorenson H.263 (.flv) - 240p>,
 <Video: H.264 (.flv) - 360p>,
 <Video: H.264 (.flv) - 480p>,
 <Video: H.264 (.mp4) - 360p>,
 <Video: H.264 (.mp4) - 720p>,
 <Video: VP8 (.webm) - 360p>,
 <Video: VP8 (.webm) - 480p>,
 <Video: VP8 (.webm) - 720p>]
>>> yt.get_highest_res('webm')
<Video: VP8 (.webm) - 720p>
>>> yt.get_highest_res()
<Video: H.264 (.mp4) - 720p>
>>> yt.get_highest_res('flv')
<Video: H.264 (.mp4) - 720p>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,13 @@ video = yt.get('mp4')
 # In this case, we'll need to specify both the codec (mp4) and resolution
 # (either 360p or 720p).
 
+# We can also get the highest resolution available using the 
+# get_highest_quality() method. An optional preferred filetype extension can be 
+# specified as a string, otherwise we default to 'mp4', if available. The function
+# will favour the highest quality over the preferred filetype, though.
+
+video = yt.get_highest_res(preferred='mp4')
+
 # Okay, let's download it!
 video.download()
 
--- a/README.md
+++ b/README.md
@@ -115,8 +115,7 @@ video = yt.get('mp4')
 # We can also get the highest resolution available using the 
 # get_highest_quality() method. An optional preferred filetype extension can be 
 # specified as a string, otherwise we default to 'mp4', if available. The function
-# will favour the highest quality over the preferred filetype, though.
-
+# will favour the highest quality over the preferred filetype.
 video = yt.get_highest_res(preferred='mp4')
 
 # Okay, let's download it!
--- a/pytube/__init__.py
+++ b/pytube/__init__.py
@@ -196,6 +196,26 @@ class YouTube(object):
             raise MultipleObjectsReturned("get() returned more than one "
                                           "object -- it returned %d!" % d)
 
+    def get_highest_res(self, preferred='mp4'):
+        """
+        Return the highest quality resolution available that matches the 
+        preferred filetype, if available. Will favour the highest quality 
+        over the preferred filetype. If not available, the highest 
+        quality that matches any filetype is returned.
+
+        Keyword arguments:
+        preferred -- the preferred extension (e.g.: mp4) 
+        """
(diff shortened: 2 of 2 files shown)


nficano:
Hi Lex,

I'm happy to see my little project is getting a some attention, and I truly
appreciate your feedback.

Let me preface this with mentioning my rationale for the project was to
build a lightweight package that makes the downloading portion of scripted
Youtube video downloads easier and not a command line utility for
downloading YouTube videos.

With that said, my appertain to this feature is it tends to push the
library towards a utility.

The way I see it is your application should import pytube, and pytube
provides you the toolset to streamline the access to information your
application needs to determine which video to download, hell maybe not even
download a video, perhaps just scrape for some specific metadata regarding
a video, who knows, but get_highest_res(self, preferred='mp4') implies an
intended use-case.

I would probably be okay with either a get_highest_res method, where
preferred is a required argument, or better yet, modifying the get() method
to specify resolution="highest" (I don't exactly like that pattern of using
string evaluation inside get() to check if resolution equals some value,
but I think you get the point).

I would eventually like to package it with a command line utility that
would install to /usr/bin - that supports completion, in which WOULD be a
good place for a --get-max-resolution-available.

Also, I want to thank you, writing this made me reflect on the project
rationale, I was reminded of (1) all the meta data contained in the http
response that I don't expose - that I probably should throw pprint to see
it yourself (
https://github.com/NFicano/python-youtube-download/blob/master/pytube/__init__.py#L167)
(2) the printing of the download status kind of alludes to an intended
use-case, so I may rethink the way the way downloads get handled.

Anyway thanks,

feel free to write back, keep in touch, etc.

Best,
Nick

lextoumbourou:
Hey Nick,

No problem at all. I understand completely your issues with my commit.

The way I see it, a lot of scripts that could potentially use your library - especially ones that are scraping Youtube for content in some way - are going to want a way to handle the uncertainties of available file types and resolutions. And, obviously, in most cases, the developer of said script is going to want the script to go for the highest available quality.

So, I think your suggestion to modify the behavior of the get() method does indeed sound like the cleanest idea. With that said, here are another couple of ideas: 
1. We could add an optional bool argument to the get() method like: highest=True. When it's set, resolution is ignored and we just shoot for the highest. This sounds the most reasonable to me, as it won't break scripts currently utilising your library.
2. Or, what about if we modified the get() method so that if a resolution isn't specified, we look for the highest by default instead of raising a MultipleObjectsReturned expection? I had a big night last night, so I'm probably missing something obvious, but I think it would be most useful to script writers if by default, we guaranteed something would be returned if there was video content available.

Either way, thanks for the library :)

Lex

ludwich:
i would love to see this feature implemented

Issue #6, opened 2013-01-17
module error


ludwich:
i get an error on every 2-3 run.
http://www.youtube.com/watch?v=zLTOzu8N1Pg&feature=youtube_gdata
[<Video: MPEG-4 Visual (.3gp) - 144p>, <Video: MPEG-4 Visual (.3gp) - 240p>, <Video: Sorenson H.263 (.flv) - 240p>, <Video: H.264 (.flv) - 360p>, <Video: H.264 (.flv) - 480p>, <Video: H.264 (.mp4) - 360p>, <Video: H.264 (.mp4) - 720p>, <Video: VP8 (.webm) - 360p>, <Video: VP8 (.webm) - 480p>, <Video: VP8 (.webm) - 720p>]
Traceback (most recent call last):
  File "test2.py", line 40, in <module>
    video.download('/c/media')
  File "build/bdist.linux-x86_64/egg/pytube/models.py", line 36, in download
  File "/usr/lib/python2.7/urllib2.py", line 127, in urlopen
    return _opener.open(url, data, timeout)
  File "/usr/lib/python2.7/urllib2.py", line 407, in open
    response = meth(req, response)
  File "/usr/lib/python2.7/urllib2.py", line 520, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib/python2.7/urllib2.py", line 445, in error
    return self._call_chain(_args)
  File "/usr/lib/python2.7/urllib2.py", line 379, in _call_chain
    result = func(_args)
  File "/usr/lib/python2.7/urllib2.py", line 528, in http_error_default
    raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
urllib2.HTTPError: HTTP Error 403: Forbidden

rbrcsk:
I can confirm this. I've done some testing:

Here's a link that results in a 403: 

```
[...]

Issue #5, opened 2013-01-17
Index out of range 


ludwich:
http://www.youtube.com/watch?v=zLTOzu8N1Pg&feature=youtube_gdata
Traceback (most recent call last):
  File "test2.py", line 35, in <module>
    yt.url = data['feed']['entry'][0]['link'][0]['href']
  File "build/bdist.linux-x86_64/egg/pytube/api.py", line 75, in url
  File "build/bdist.linux-x86_64/egg/pytube/api.py", line 211, in _get_video_info
IndexError: list index out of range

hansihe:
My issue does seem to be the same as this one, so I will provide some more information.
This error only seems to appear occasionally. I would say about 1/10 of attempts.
Here is my full trace:

```
Traceback (most recent call last):
  File "C:\Users\HansiHE\PycharmProjects\YouTubeCards\run.py", line 10, in <module>
    youtube.url = 'http://www.youtube.com/watch?v=kahpg789lGs'
  File "C:\Users\HansiHE\PycharmProjects\YouTubeCards\pytube\api.py", line 74, in url
    self._get_video_info()
  File "C:\Users\HansiHE\PycharmProjects\YouTubeCards\pytube\api.py", line 210, in _get_video_info
    signature = video_signatures[idx]
IndexError: index out of range: 8
```

It occurs on this piece of code:

``` python
from pytube.api import YouTube

youtube = YouTube()
youtube.url = 'http://www.youtube.com/watch?v=kahpg789lGs'
print youtube.filter('mp4', '360p')
```

rbrcsk:
@hansihe I described what's wrong on issue #6. The same error is causing this too.

hansihe:
@randomchars Ah, thanks for letting me know.

heetbeet:
This error seems to persist
everything runs fine with:

```
yt = YouTube()
yt.url = 'http://www.youtube.com/watch?v=kahpg789lGs'
```

but there is a problem with:

```
yt = YouTube()
yt.url = 'https://www.youtube.com/watch?v=PIb6AZdTr-A'
```

```
IndexError                                Traceback (most recent call last)
<ipython-input-8-a8b54bd239d1> in <module>()
----> 1 yt.url = 'https://www.youtube.com/watch?v=PIb6AZdTr-A'

C:\Users\Simon\Dropbox\petprojects\youtube_downloader\pytube\api.pyc in url(self, url)
     74         self._filename = None
     75         #Get the video details.
[...]

PR #7, opened 2013-01-18
Video info parsing and Python 3 fixes

rbrcsk:
[This](https://github.com/NFicano/python-youtube-download/commit/f9849241120e9bec548f8d7113409d2be6291a0c) commit broke compatibility with Python 3. In line  41 `self._buffer` is set to `buffer` which doesn't exists in 3.X.

The equivalent would be `memoryview` which works in 2.7+ but seeing as `buffer` was not used at all, (It was overridden two lines below) I just removed it.

### DIFF
--- a/pytube/api.py
+++ b/pytube/api.py
@@ -5,7 +5,7 @@ from .models import Video
 from .utils import safe_filename
 from urllib import urlencode
 from urllib2 import urlopen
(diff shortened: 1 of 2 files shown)


rbrcsk:
Updated with a fix for #5 and #6 .

PR #10, opened 2013-01-30
Add a comman line interface

rbrcsk:
As the title says, I've added a command line interface. It doesn't have too many options, but it's a start.


nficano:
@randomchars you sir, are fucking awesome.

PR #27, opened 2014-08-16
added a progress bar and percentage of download completed along with human readable file size

sanfx:
added a progress bar and percentage of download completed along with human readable file size.

Example image

### DIFF
--- a/pytube/models.py
+++ b/pytube/models.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 from os.path import normpath, isfile
 from os import remove
+from time import clock
 try:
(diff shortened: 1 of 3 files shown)


sanfx:
Just updated a bit more(later that day I discovered printing on windows doesn't work fine so that is fixed along with few extra things added) convert relative path to absolute path at the time being passed from commad line, updated Downlaoding message to print where the file is saved to, check to make sure user passes both extension and resolution and print to shell if not sure what resolution to choose, and fix for progress bar dispay on MS Windows xp - 8.

Updated shell display example

PR #41, opened 2014-12-08
Added posibility to save movie to custom file name

Eloar:


josegonzalez:
You should redo this PR, as it appears to have a bunch of things not related to the pr.

PR #45, opened 2014-12-10
raising exception instead of application exit on error

Eloar:
- changed all exit usages to raise exceptions


josegonzalez:
Can't be automerged, can you redo the commit?

Eloar:
merged master to 4_issue branch resolving conflict so it should work now

PR #52, opened 2015-04-10
GUI

parajain:
It will be great with gui.


nficano:
I don't want to add any dependances. Its one of the main goals of the project.

PR #71, opened 2015-10-04
Unittests for YouTube class

garg10may:
Covered public methods. Testing for http://www.youtube.com/watch?v=Ik-RsDGPI5Y, which is part of docuentation.


nficano:
GREAT!!!!

Issue #88, opened 2016-01-18
getting error : cipher changed by youtube


DefCon-007:
Error Log :
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/pytube/api.py", line 368, in _get_cipher
    matches = reg_exp.search(self._js_cache)
TypeError: cannot use a string pattern on a bytes-like object

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/Ayush/Documents/CODE/GIT_youtube/Project-Youtube/YouTube_Search.py", line 174, in <module>
    download_video_via_url(video_link[int(serial_no[a])-1])
  File "/Users/Ayush/Documents/CODE/GIT_youtube/Project-Youtube/YouTube_Search.py", line 113, in download_video_via_url
    yt=YouTube(video_url)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/pytube/api.py", line 66, in **init**
    self.from_url(url)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/pytube/api.py", line 189, in from_url
    signature = self._get_cipher(stream_map["s"][idx], js_url)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/pytube/api.py", line 379, in _get_cipher
    "issue on GitHub: {0}".format(e))
pytube.exceptions.CipherError: Couldn't cipher the signature. Maybe YouTube has changed the cipher algorithm. Notify this issue on GitHub: cannot use a string pattern on a bytes-like object

nficano:
Can you please give me the video url so I can try to repro this? thanks!

DefCon-007:
Here is the url :     https://www.youtube.com/watch?v=rClUOdS5Zyw   I guess
the problem is with the vevo channels.

icemanovich:
Other video link example with the same bug
https://www.youtube.com/watch?v=gfjhqx3bbdY

nficano:
Can @ayush91195 and @icemanovich give me your Python and pytube versions? Both videos seem to work fine under Python 2.7.11 and pytube 6.1.5.

``` python
>>> from pytube import YouTube
>>> yt = YouTube('https://www.youtube.com/watch?v=rClUOdS5Zyw')
>>> yt.get_videos()
[...]

PR #90, opened 2016-01-29
fixed issue #88

kmad1729:
Hi,
The issue is happening because of str and bytes difference in python3.
re.search wouldn't work on bytes-like object.


nficano:
Thank you sir!

PR #126, opened 2017-03-27
Simplified sizeof utility

DonoA:
Shortens the sizeof util to be more concise and efficient. Renames the parameter for sizeof so it does not overwrite the `bytes` function.

### DIFF
--- a/pytube/utils.py
+++ b/pytube/utils.py
@@ -50,7 +50,9 @@ def sizeof(byts):
     """
     sizes = ['bytes', 'KB', 'MB', 'GB', 'TB', 'PB']
     power = int(math.floor(math.log(byts, 1024)))
-    return str(int(byts/float(1024**power)))+' '+(sizes[power] if byts != 1 else 'byte')
+    value = int(byts/float(1024**power))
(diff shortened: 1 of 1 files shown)


pytube/utils.py
```diff
@@ -41,34 +42,15 @@ def safe_filename(text, max_length=200):
...
-        singular, multiple = suffix
-        if amount == 1:
-            suffix = singular
-        else:
-            suffix = multiple
-    return "%s%s" % (str(amount), suffix)
-
+    sizes = ['bytes', 'KB', 'MB', 'GB', 'TB', 'PB']
+    power = int(math.floor(math.log(byts, 1024)))
+    return str(int(byts/float(1024**power)))+' '+(sizes[power] if byts != 1 else 'byte')
```
nficano: please assign this to variables and use the ``.format()`` method. See https://pyformat.info/

nficano (changes requested):

DonoA:
Sorry about that, still fairly new to python and don't know all the cool tricks. Hope this is a bit better.

nficano (approved):
LGTM

PR #127, opened 2017-04-05
if the path is not a directory raise an OSError

piratimir:
Checks if the path is a directory or not, if not raise an OSError, else continue.


piratimir:
@nficano Hey man I fixed the case when the path does not match a directory by raising an OSError. Check it :)

nficano (approved):

PR #148, opened 2017-08-26
unswap 'Resolution' and 'Extensions' in doc header + minor linting

adrianstaniec:


adrianstaniec:
no comment?

nficano:
sorry, that was quite rude, wasn't it. I rewrote pytube from scratch, so this is no longer compatible as the CLI now dumps out more information about each stream

PR #162, opened 2017-10-17
Fix CLI

vvuk:
The CLI seems to be using an old version of the get() method and progress handler... this fixes it to actually work when fetching a stream.


nficano (approved):

nficano:
Thanks a lot @vvuk!

PR #186, opened 2017-12-30
fix by removing " when entered pathname has "

AnkanDas22:
Many a times, the pathname can have " when path is copied by "Copy as Path" in Windows or even in other circumstances.
To avoid errors arising from " altogether, I have added code to remove " from entered pathname (i.e. destination path).


nficano:
there's an easier way to do this: https://github.com/nficano/pytube/commit/83be388c5657c58cfc2cf66ad9f449d5266fe525

PR #187, opened 2018-01-08
Update __main__.py

lvaleriu:
Check if key is in dictionary


lvaleriu:
Check if key is in dictionary

PR #226, opened 2018-02-20
Choose playlist download location

ahmedsadman:
Feature added for choosing playlist download location


jackson-sandland:
can you only set the download location for a playlist?
would be great to set download location for a single file

PR #230, opened 2018-03-01
Fixes #224

### LINKED ISSUE #224: Downloading playlist fails on age restricted videos
self.js_url and self.js variable not set when downloading age_restricted videos, which cause an exception when line 40 of cipher.py is hit because js is None.

billd100:
#224


billd100:
@nficano Please let me know what I need to do to get this PR in order.

Travis error looks to be Flake8 related.

nficano (approved):

PR #232, opened 2018-03-02
Feature: cache contents of base.js from identical urls

billd100:
YouTube appears to serve base.js from the same URL pretty often. This should cut down on the number of requests we're making for the same resource when downloading more than a single video.


nficano:
solution is incompatible w/ python 2.7.

PR #248, opened 2018-04-04
Stream video to buffer

jkuijt:
I've added a function to stream video data to a buffer. 
`Stream.stream_to_buffer()` will save all video data in an `io.BytesIO` object instead of a file. 
I find this useful when I do post processing on videos.


nficano:
perfect contribution! 👏

nficano (approved):

PR #253, opened 2018-04-07
added autonumbering for playlists

Jan-V2:
Added optional automatic numbering of the files created by Playlisl.download_all.

This is done by using a generator function that yields a filename prefix, which is passed to Stream.download.

### DIFF
--- a/pytube/contrib/playlist.py
+++ b/pytube/contrib/playlist.py
@@ -66,12 +66,25 @@ class Playlist(object):
             complete_url = base_url + video_id
             self.video_urls.append(complete_url)
 
-    def download_all(self, download_path=None, prefix_number=True):
+    def download_all(self, download_path=None, prefix_number=True, reverse_numbering=False):
         """Download all the videos in the the playlist. Initially, download
         resolution is 720p (or highest available), later more option
         should be added to download resolution of choice
 
         TODO(nficano): Add option to download resolution of user's choice
+
+        :param download_path:
+            (optional) Output path for the playlist If one is not
+            specified, defaults to the current working directory.
+            This is passed along to the Stream objects.
+        :type download_path: str or None
+        :param prefix_number:
+            (optional) Automatically numbers
+        :type prefix_number: bool
+        :param reverse_numbering:
+            (optional) Lets you number playlists in reverse, since some
+            playlists are ordered newest -> oldests.
+        :type reverse_numbering: bool
         """
 
         def path_num_prefix_generator():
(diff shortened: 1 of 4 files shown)


Jan-V2:
sorry about all these commits with linting errors, this is my first pr.
i've now figured out how to run flake om by local machine, so that should be sorted from now on.

pytube/contrib/playlist.py
```diff
@@ -66,18 +66,64 @@ def populate_video_urls(self):
...
+        This generator function generates number prefixes, for the items
+        in the playlist.
+        If the number of digits required to name a file,is less than is
+        required to name the last file,it prepends 0s.
+        So if you have a playlist of 100 videos it will number them like:
+        001, 002, 003 ect, up to 100.
+        It also adds a space after the number.
+        :return: prefix: string
+        """
+        digits = len(str(len(self.video_urls)))
```
bartlomiejwiater: @johnvanderholt Do you think it will do the job?
```python
def _path_num_prefix_generator(video_urls, reverse=False):
    digits = len(str(len(video_urls)))
    start, stop, step = (len(video_urls), 0, -1) if reverse else (1, len(video_urls) + 1, 1)
    return (str(i).zfill(digits) for i in range(start, stop, step))
```

Just add `self` parameter and use `self.video_urls`.


You can use it with `print(list(_path_num_prefix_generator(['s'] * 78)))` or `print(list(_path_num_prefix_generator(['s'] * 178, reverse=True)))`.

Jan-V2: looks great!
i've committed it to the pr, but i did redactor it a bit to make it, in my opinion, more readable.
```python
digits = len(str(len(self.video_urls)))
        if reverse:
            start, stop, step = (len(self.video_urls), 0, -1)
        else:
            start, stop, step = (1, len(self.video_urls) + 1, 1)
        return (str(i).zfill(digits) for i in range(start, stop, step))
```

Jan-V2:
so the project won't build anymore on python 3.4. it seems to be a bug related to the pipfile.lock the thing is i didn't change the pipfile.lock between now and my last working commit. 
could someone more familair with the project like @nficano  take a closer look at this.

Jan-V2:
this is the error i got:

`
THESE PACKAGES DO NOT MATCH THE HASHES FROM Pipfile.lock!. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    pluggy==0.6.0 from https://files.pythonhosted.org/packages/ba/65/ded3bc40bbf8d887f262f150fbe1ae6637765b5c9534bd55690ed2c0b0f7/pluggy-0.6.0-py3-none-any.whl#sha256=e160a7fcf25762bb60efc7e171d4497ff1d8d2d75a3d0df7a21b76821ecbf5c5 (from -r /tmp/pipenv-1lhacawa-requirements/pipenv-qfd5nzwg-requirement.txt (line 1)):
`

so it appears that the pluggy package that it downloaded has a different hash than was in the Pipfile.lock. which seems really weird since it's the exact same version that it used in working builds like here https://travis-ci.org/nficano/pytube/jobs/365242847.
still weird though that it's only happening in python 3.4 though.
on top of that i checked the master branch commits and it appears that nobody else touched Pipfile.lock either. 
basically i'm not quite sure what the problem is, but it seems to be related to the package hash.

nficano (approved):

PR #255, opened 2018-04-10
Add proxy support.

BenZstory:
- add proxy option for YouTube, install a global proxy for urllib.
- compatible with both py27 urllib2 and py35 urllib
- example: yt = YouTube(url, proxies={"http":"http://my.proxy.com:1234"})


nficano (approved):

nficano:
great contribution! thanks a lot!

PR #264, opened 2018-04-20
Return downloaded file path at the end of download

infopz:
It's useful to get the downloaded file path at the end of a download, if a user have to do something with that file


grking8:
filepath can be inferred from the defaults I believe

nficano (approved):

PR #267, opened 2018-05-07
added the ability the choose the resolution at which you download a playlist

Jan-V2:
added the ability the choose the resolution at which you download a playlist, and fixed minor bug with playlist numbering (was missing a space).


hbmartin:
@johnvanderholt added this functionality to pytube3: https://github.com/hbmartin/pytube3

RONNCC:

PR #271, opened 2018-05-10
add option to skip existing files previously downloaded

fja05680:
looks in local download path for same filename and file size.  Skip the download and overwrite of the local file if skip_existing=True


hbmartin:
@fja05680 opened a PR to add this functionality to pytube3, could you take a look? https://github.com/hbmartin/pytube3/pull/25

RONNCC:

Issue #252, opened 2018-04-06
Ordering by 'abr' orders alphabetically, not numerically


ndg63276:
Ordering by 'abr', gives orders as if sorted alphabetically, so 128 < 160 < 50 < 70, eg

`>>> yt=YouTube('https://www.youtube.com/watch?v=9bZkp7q19f0')`
`>>> yt.streams.filter(only_audio=True).order_by('abr').all()`
`[<Stream: itag="140" mime_type="audio/mp4" abr="128kbps" acodec="mp4a.40.2">, <Stream: itag="171" mime_type="audio/webm" abr="128kbps" acodec="vorbis">, <Stream: itag="251" mime_type="audio/webm" abr="160kbps" acodec="opus">, <Stream: itag="249" mime_type="audio/webm" abr="50kbps" acodec="opus">, <Stream: itag="250" mime_type="audio/webm" abr="70kbps" acodec="opus">]`

Is it possible to do a natural sort, so that 50 < 70 < 128 < 160?

Mattwmaster58:
formatted
```
>>> yt=YouTube('https://www.youtube.com/watch?v=9bZkp7q19f0')
>>> yt.streams.filter(only_audio=True).order_by('abr').all()
[<Stream: itag="140" mime_type="audio/mp4" abr="128kbps" acodec="mp4a.40.2">, 
 <Stream: itag="171" mime_type="audio/webm" abr="128kbps" acodec="vorbis">, 
 <Stream: itag="251" mime_type="audio/webm" abr="160kbps" acodec="opus">,
 <Stream: itag="249" mime_type="audio/webm" abr="50kbps" acodec="opus">,
 <Stream: itag="250" mime_type="audio/webm" abr="70kbps" acodec="opus">]
```
The `order_by` just runs everything through the `sorted` function with `key` set to `lambda thingtosortby: getattr(self, thingtosortby)`.

Mattwmaster58:
Related to #194

vn-ki:
Closed in #287

PR #287, opened 2018-06-23
fix numeric sorting

Mattwmaster58:
Should fix #286 and #252

### DIFF
--- a/pytube/query.py
+++ b/pytube/query.py
@@ -171,7 +171,7 @@ class StreamQuery:
             integer_attr_repr[attr] = int(''.join(num)) if num else None
 
         # if every attribute has an integer representation
-        if all(integer_attr_repr.values()) and integer_attr_repr:
+        if integer_attr_repr and all(integer_attr_repr.values()):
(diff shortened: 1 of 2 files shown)


pytube/query.py
```diff
@@ -162,9 +162,23 @@ def order_by(self, attribute_name):
...
+        integer_attr_repr = {}
+        for stream in self.fmt_streams:
+            attr = getattr(stream, attribute_name)
+            if attr is None:
+                break
+            num = [char for char in attr if char.isdigit()]
+            integer_attr_repr[attr] = int(''.join(num)) if num else None
+
+        # if every attribute has an integer representation
+        if all(integer_attr_repr.values()) and integer_attr_repr:
```
vn-ki: Shouldn't this be `integer_attr_repr and all(integer_attr_repr.values())`?

Mattwmaster58: just curious, is short circuiting the only reason? readablilty? either way I'll make the change

vn-ki (approved):
Thank you @Mattwmaster58

PR #288, opened 2018-06-23
fix streams.get_by_itag function.

### LINKED ISSUE #280: get_by_itag takes int, not string
The [documentation of get_by_itag for v9.2.2](http://python-pytube.readthedocs.io/en/latest/user/quickstart.html#get-streams-by-itag) is to pass in the desired stream's itag value as a srting. However the the StreamQuery class builds the itag_index dictionary with the keys as integers (line 15 in query.py). So when get_by_itag (line 187 in query.py) is called with a string parameter, the look up in itag_index fails and returns None.

```   
  yt = pytube.YouTube('https://www.youtube.com/watch?v=XaGHcBZjmWA')

  for yt_stream in yt.streams.filter(only_audio=True).all():
        print('{}'.format(yt_stream.itag))
        print(yt.streams.get_by_itag(yt_stream.itag))
```

Will print out 

```
140
None
170
None
(etc)
```

This could be a documentation fix or change the mapping in query depending on if  itag is intended be treated either an integer or string.

Mattwmaster58:
Should fix #280


vn-ki (changes requested):
Why not support both `int` and `str`? We don't want to break backwards compatibility. The travis failure is due to that.

vn-ki (approved):

PR #290, opened 2018-06-24
Cache filesize

### LINKED ISSUE #212: Shouldn't filesize be a cached property?
https://github.com/nficano/pytube/blob/c052b5cb7d6a0280e46dbd2c495a1995b78e27c7/pytube/streams.py#L153-L162

Apparently it conducts requests on every property call. Note that in python even `hasattr` calls property code.

Mattwmaster58:
Cache the `filesize` so that it is only requested for once. Fixes #212 

PS
those first four commits are just me not gitting it.


vn-ki (approved):
Looks ok. Thanks for the PR!

PR #297, opened 2018-07-22
Remove test downloads

grking8:


grking8:
`test_playlist.py` removed - tests now faster

PR #305, opened 2018-08-13
Save one network query when video not age restricted

apiraino:
Hi,

I think here we have a small but free optimization; we can shave off `GET`ing the embed HTML if the video is not age restricted.

Does it make sense?


pytube/__main__.py
```diff
@@ -160,8 +160,9 @@ def prefetch(self):
         self.watch_html = request.get(url=self.watch_url)
         if '<img class="icon meh" src="/yts/img' not in self.watch_html:
             raise VideoUnavailable('This video is unavailable.')
```
apiraino, line 162: Also, I don't quite understand the logic of this `if` statement. Maybe should it be inverted?

r? @nficano

apiraino:
hi @nficano reminder about this

RONNCC:

PR #320, opened 2018-09-30
Suppress exception while downloading Playlist

sonicaj:
This commit adds an option to suppress exception when downloading a video in a playlist ( i.e Video Unavailable ). This way rest of the playlist still continues to be downloaded instead of breaking up in the middle of the download process

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ _templates
 _autosummary
 
 .pytest_cache*
+
+# IDE Files
+.idea/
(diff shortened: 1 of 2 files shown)


sonicaj:
@nficano do you think we should envelope the whole downloading block in the try catch in case a downloading stream fails whereas the video is still available by youtube ?

nficano:
that's bad practice, if there's an error, I would rather it bubble up as an issue over silently ignoring it.

sonicaj:
A case: I am downloading 500 videos which are in a playlist and the 200th video is unavailable. The code crashes and downloading is disrupted. There is no way that I can get rest of the 300 videos.
I think this hurts users experience incredibly and I do agree with you but this case needs to be accounted for. Thoughts ?

nficano (approved):

Issue #333, opened 2018-11-07
getting regex error


proxprox:
Upon downloading a playlist, i get the regex error 
`Exception in Tkinter callback
Traceback (most recent call last):
  File "C:\Python34\lib\tkinter\__init__.py", line 1487, in __call__
    return self.func(*args)
  File "C:/Users/nicto/PycharmProjects/idk/playlist activator/playlist downloader.py", line 40, in func
    pl.download_all('C:\\pythonpractice\music')
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\contrib\playlist.py", line 150, in download_all
    yt = YouTube(link)
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\__main__.py", line 88, in __init__
    self.prefetch_init()
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\__main__.py", line 97, in prefetch_init
    self.init()
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\__main__.py", line 133, in init
    mixins.apply_signature(self.player_config_args, fmt, self.js)
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\mixins.py", line 49, in apply_signature
    signature = cipher.get_signature(js, stream['s'])
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\cipher.py", line 250, in get_signature
    tplan = get_transform_plan(js)
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\cipher.py", line 69, in get_transform_plan
    name = re.escape(get_initial_function_name(js))
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\cipher.py", line 45, in get_initial_function_name
    return regex_search(pattern, js, group=1)
  File "C:\Users\nicto\PycharmProjects\idk\venv\lib\site-packages\pytube\helpers.py", line 41, in regex_search
    .format(pattern=p),
pytube.exceptions.RegexMatchError: regex pattern (yt\.akamaized\.net/\)\s*\|\|\s*.*?\s*c\s*&&\s*d\.set\([^,]+\s*,\s*(?P<sig>[a-zA-Z0-9$]+)\() had zero matches`

ytensor42:
yeah, I am experiencing same error since few hours ago.
[...]

PR #334, opened 2018-11-07
Update regex for locating signature function.

billd100:
Addresses issue #333.


billd100:
@nficano Hey, I know you cut a release earlier today but when can we expect one with this update?

PR #340, opened 2018-11-18
 Replace mocks with newly fetched data and fix some assertion

kurokobo:
For meaningful unit tests, replace mocks with newly fetched data and fix assertions with real data which has fetched online. Now all test has passed correctly.
In addition, for follow PEP-8, insert some line break to make the line shorter than 79 chars.


kurokobo:
Sorry, I missed some indent in `__main__.py` and `cipher.py`... :(
I will fix it and send again later, sorry for taking your time.

PR #345, opened 2018-12-03
Update playlist.py

unic0rned:
Hey,

I noticed when using the playlist function on a big german playlist, that it would stop on URLS with % signs in it.
But instead of stopping the whole program, I would prefer that it just continues to download the rest of the files.
Somehow this fix also allowed me to download the files even with % signs in them.
Before it would break with saying NoneType object has no download function.

### DIFF
--- a/pytube/contrib/playlist.py
+++ b/pytube/contrib/playlist.py
@@ -165,11 +165,14 @@ class Playlist(object):
                     progressive=True, subtype='mp4',
                 ).order_by('resolution').desc().first()
 
-                logger.debug('download path: %s', download_path)
-                if prefix_number:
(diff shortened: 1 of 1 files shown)


hbmartin:
@unic0rned do you have an example playlist URL that shows this behavior?

unic0rned:
No Sorry I don't have one

Am Di., 21. Jan. 2020 um 19:42 Uhr schrieb Harold Martin <
notifications@github.com>:

hbmartin:
@unic0rned this is fixed in pytube3: https://github.com/hbmartin/pytube3

RONNCC:

PR #348, opened 2018-12-22
Fixed comment in __main__.py

viagluck:
Corrected the name of a param in a comment to reflect the actual name found in the code.


hbmartin:
@viagluck  fixed in my Python 3 only fork: https://github.com/hbmartin/pytube3

RONNCC:

PR #352, opened 2019-01-06
possible fix #347: added check for ciphered signature existence in stream

### LINKED ISSUE #347: Key Error in mixing.py line 49
I got this error once every some videos. The weird thing is that for the same url it does not happen every time. Here is a screen shot of the error

screen shot 2018-12-20 at 10 47 26 am

mitrabeast:
There was no check for `'s'` in `stream`, so that was able to cause unexpected KeyError exceptions. I've added check but didn't know ignore it or not, or use some type of exception in this case. I've added TypeError one, but I think it should be processed anyway better.

### DIFF
--- a/pytube/mixins.py
+++ b/pytube/mixins.py
@@ -46,7 +46,12 @@ def apply_signature(config_args, fmt, js):
             continue
 
         if js is not None:
-            signature = cipher.get_signature(js, stream['s'])
+            if 's' not in stream:
+                logger.error('Ciphered signature not found in stream!')
+                # Can be changed to another exception typem or simply ignored.
+                raise TypeError('Ciphered signature is None')
+            ciphered_signature = stream['s']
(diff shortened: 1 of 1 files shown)


Choromanski:
+1

topplethepat:
when I added this I got the error: 

        pytube.exceptions.RegexMatchError: regex pattern (;ytplayer\.config\s*=\s*({.*?});) had zero matches

earamburo:
@topplethepat  I am also getting this error. How do I fix it?

irobot-zyj:
> @topplethepat I am also getting this error. How do I fix it?

Omiting the last field t=0s works for me.

hbmartin:
@masta2898 @earamburo what video URLs required this change?

RONNCC:
Hi @m1tr4b34st 

Thanks

PR #370, opened 2019-03-17
Change regex

grking8:
- for https://github.com/nficano/pytube/issues/368, can now download https://www.youtube.com/watch?v=Ck2jYtY5Hgg
- regex taken from https://github.com/ytdl-org/youtube-dl/blob/77d95677b7ab4a9840ef142b14627b07a9a31120/youtube_dl/extractor/youtube.py#L1570
- CI pipeline failing due to https://pypi.org/project/more-itertools/ (see version history) no longer supporting python 2.7?

### DIFF
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -47,7 +47,7 @@ def is_age_restricted(watch_html):
         Whether or not the content is age restricted.
     """
     try:
-        regex_search(r'og:restrictions:age', watch_html, group=0)
+        regex_search(r'player-age-gate-content">', watch_html, group=0)
     except RegexMatchError:
         return False
     return True


stdedos:
:expressionless: it didn't occur to me to check if the video was age-restricted (although I generally browse ingognito-mode)

Maybe, however, both versions of the regex are needed?

It may not be straightforward which version is used where.

grking8:
I found browsing in incognito in chrome I was able to view  https://www.youtube.com/watch?v=Ck2jYtY5Hgg whereas https://www.youtube.com/watch?v=irauhITDrsE prompted me to sign in

which is why I found `og:restrictions:age` needed to be **replaced** by `player-age-gate-content">` (the first url html contains `og:restrictions:age` (and not `player-age-gate-content`) but should have `is_age_restricted(watch_html)` return `False`, only those urls prompting  user for login should return `True`)

nficano (approved):

RONNCC:
Hi @grking8 

Thanks

PR #375, opened 2019-03-27
Added functionality to only parse a set amount of links

Choromanski:
Allows parsing of partial playlist

### DIFF
--- a/pytube/contrib/playlist.py
+++ b/pytube/contrib/playlist.py
@@ -52,7 +52,7 @@ class Playlist(object):
             load_more_url = ''
         return load_more_url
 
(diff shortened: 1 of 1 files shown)


hbmartin:
@Choromanski are there any tests for this?

Choromanski:
Honestly not even sure if this works anymore.
I'll look into it and make up some tests. Hopefully some time this weekend.

hbmartin:
I made a PR for pytube3 that does almost the same thing https://github.com/hbmartin/pytube3/pull/26 except using a video id instead of an index

RONNCC:

PR #383, opened 2019-04-29
Temporary fix for  regex pattern error

manuhg:
Temporary fix for  regex pattern (\W[\'"]?t[\'"]?: ?[\'"](.+?)[\'"]) had zero matches


RONNCC:

Issue #380, opened 2019-04-18
[Question?] How to change video name save ?


hangocanh2303:
I want to change name of video when I save it, how can i do that ?

Medstar117:
You answer lies [here](https://python-pytube.readthedocs.io/en/latest/api.html#pytube.Stream.download) in the pytube API. One of the download function’s optional parameters is as follows:

> filename (str or None) – (optional) Output filename (stem only) for writing media file. If one is not specified, the default filename is used.

“Stem only“ refers to just the name you want, not including the file extension (e.g. filename=“example”, **NOT** filename=“example.mp4”). If you are using the CLI rather than the Python IDLE, then there is currently no way to specify neither a name nor a download target for the video stream (although @nficano has made a TODO comment within the cli.py script for such a parameter).

For future reference (not trying to sound like a smartass, just trying to be helpful as to avoid future issues with other people), this is a question, not an issue. Issues are related to errors within the program or to code snippets no longer working as intended, and are to be posted in this section of the repository for people to be notified and, hopefully, for people to provide a solution. If you have a question, such as this, I would recommend posting it on sites such as [StackOverflow](https://www.stackoverflow.com/) rather than posting it here. Thank you~ :)

harishbp:
yt.streams.first().download(output_path="path_to_dir", filename="filename")

egemenayd:
Hi. I have a question. I want to name the videos by input. I tried to use inputs inside like ;   

```
"what would you like to do file name? "
input= name 

yt.download("D:/yt", name) 

```
(i know syntax is wrong but I tried to explain I hope I did..:)

Issue #362, opened 2019-02-26
Download above 720


billapto:
Hi all,
 i try to download a 2k video via cli with pytube (on windows 10) 
like that pytube https://www.youtube.com/watch?v=nBYZpsbu9ds --itag=313
and i keep getting errors 

Transient - 4K UHD 1000FPS.webm | 255892305 bytes
'stty' is not recognized as an internal or external command,
operable program or batch file.
Traceback (most recent call last):
  File "c:\program files\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\program files\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Program Files\Python37\Scripts\pytube.exe\__main__.py", line 9, in <module>
  File "c:\program files\python37\lib\site-packages\pytube\cli.py", line 64, in main
    download(args.url, args.itag)
  File "c:\program files\python37\lib\site-packages\pytube\cli.py", line 169, in download
    stream.download()
  File "c:\program files\python37\lib\site-packages\pytube\streams.py", line 228, in download
    self.on_progress(chunk, fh, bytes_remaining)
  File "c:\program files\python37\lib\site-packages\pytube\streams.py", line 285, in on_progress
    on_progress(self, chunk, file_handler, bytes_remaining)
  File "c:\program files\python37\lib\site-packages\pytube\cli.py", line 148, in on_progress
    display_progress_bar(bytes_received, filesize)
  File "c:\program files\python37\lib\site-packages\pytube\cli.py", line 121, in display_progress_bar
    _, columns = get_terminal_size()
  File "c:\program files\python37\lib\site-packages\pytube\cli.py", line 98, in get_terminal_size
    rows, columns = os.popen('stty size', 'r').read().split()
ValueError: not enough values to unpack (expected 2, got 0)

Can anyone help?

regards

Medstar117:
You’re on Windows, I take it? I’ve put up a PR recently that fixes that. Pytube is calling for a bash command that isn’t valid on Windows in order to get the terminal’s geometry. I’ll reference your issue in my PR.

jamadagni:
I found the same problem on Linux (Kubuntu Bionic) when I am piping a list of URLs to pytube so: `cat urls.txt | while read url ; do pytube --itag 22 $url ; done`. Downloading them individually isn't a problem.

Issue #202, opened 2018-01-17
Download not callable from script without tty


ocervell:
The `pytube.cli.download` function cannot be called / run in environments where tty / stty is not available:

```
Traceback (most recent call last):
  File "/root/scrapez/venv/local/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/root/scrapez/venv/local/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/root/scrapez/tasks/youtube.py", line 34, in download
    stream.download(output_path=output_path, filename=stream.default_filename)
  File "/root/scrapez/venv/local/lib/python2.7/site-packages/pytube/streams.py", line 210, in download
    self.on_progress(chunk, fh, bytes_remaining)
  File "/root/scrapez/venv/local/lib/python2.7/site-packages/pytube/streams.py", line 246, in on_progress
    on_progress(self, chunk, file_handler, bytes_remaining)
  File "/root/scrapez/tasks/youtube.py", line 129, in on_progress
    display_progress_bar(bytes_received, filesize)
  File "/root/scrapez/venv/local/lib/python2.7/site-packages/pytube/cli.py", line 121, in display_progress_bar
    _, columns = get_terminal_size()
  File "/root/scrapez/venv/local/lib/python2.7/site-packages/pytube/cli.py", line 98, in get_terminal_size
    rows, columns = os.popen('stty size', 'r').read().split()
ValueError: need more than 0 values to unpack
```

This issue is basically the same as https://github.com/tldr-pages/tldr-python-client/issues/9.
The pull request addressing this is at https://github.com/tldr-pages/tldr-python-client/pull/18

It states that 'stty' or 'tty' is not available (or that you run from inside a multiprocessing task runner like Celery).

Can we make the `display_progress_bar` conditional ? So that we can use the `download` function from scripts running on servers.

Issue #180, opened 2017-12-29
Really slow downloading with CLI vs Interactive Prompt


stdedos:
Executing the CLI takes a little over eternity

```
alkpc@alkPC-Asus ~/.installs $  pytube https://www.youtube.com/watch?v=9wcSVTErT2U --list
<Stream: itag="22" mime_type="video/mp4" res="720p" fps="30fps" vcodec="avc1.64001F" acodec="mp4a.40.2">
<Stream: itag="43" mime_type="video/webm" res="360p" fps="30fps" vcodec="vp8.0" acodec="vorbis">
<Stream: itag="18" mime_type="video/mp4" res="360p" fps="30fps" vcodec="avc1.42001E" acodec="mp4a.40.2">
<Stream: itag="36" mime_type="video/3gpp" res="240p" fps="30fps" vcodec="mp4v.20.3" acodec="mp4a.40.2">
<Stream: itag="17" mime_type="video/3gpp" res="144p" fps="30fps" vcodec="mp4v.20.3" acodec="mp4a.40.2">
<Stream: itag="136" mime_type="video/mp4" res="720p" fps="30fps" vcodec="avc1.4d401f">
<Stream: itag="247" mime_type="video/webm" res="720p" fps="30fps" vcodec="vp9">
<Stream: itag="135" mime_type="video/mp4" res="480p" fps="30fps" vcodec="avc1.4d401e">
<Stream: itag="244" mime_type="video/webm" res="480p" fps="30fps" vcodec="vp9">
<Stream: itag="134" mime_type="video/mp4" res="360p" fps="30fps" vcodec="avc1.4d401e">
<Stream: itag="243" mime_type="video/webm" res="360p" fps="30fps" vcodec="vp9">
<Stream: itag="133" mime_type="video/mp4" res="240p" fps="30fps" vcodec="avc1.4d4015">
<Stream: itag="242" mime_type="video/webm" res="240p" fps="30fps" vcodec="vp9">
<Stream: itag="160" mime_type="video/mp4" res="144p" fps="30fps" vcodec="avc1.4d400c">
<Stream: itag="278" mime_type="video/webm" res="144p" fps="30fps" vcodec="vp9">
<Stream: itag="140" mime_type="audio/mp4" abr="128kbps" acodec="mp4a.40.2">
<Stream: itag="171" mime_type="audio/webm" abr="128kbps" acodec="vorbis">
<Stream: itag="249" mime_type="audio/webm" abr="50kbps" acodec="opus">
[...]

PR #385, opened 2019-05-01
CLI: Windows Compatability and Enhancements

Medstar117:
Known issues addressed:
#380 (if user was using the CLI)
#362
#202
#180 (sort of)

Updates:
Added compatability for Windows to "get_terminal_size" as to prevent cli.py from crashing when downloading a stream.
**For Linux and OSX users**, the os.popen call for "stty size" has been replaced with a subprocess.check_output call. This tweak prevents stderr from displaying an invalid bash command error on Windows every time the progress bar is updated; because this puts the "display_progress_bar" function into a different process, this edit also has a side effect of reducing download time slightly (bonus!).
If either attempt at getting the terminal geometry fails, the default values are returned as follows in an attempt to prevent an error: rows: 80, columns: 24.

Enhancements:
Added optional arguments for a target directory and a custom name for the downloaded stream.
If either argument isn't specified, the default action occurs for each (target is current working directory and name is the stream's original name, both of which are stream.download's defaults).


Medstar117:
Oof, that's a lot of coverage loss...any help on fixing that would be greatly appreciated. :)

Also, should I split this PR into two separate ones--one for Windows coverage, and one for the new arguments? Or should I leave it as one bulk PR?

Edit:
Well, that coverage boost attempt helped a bit...it's better than -3.2%. I'm just not experienced enough to work with the mock module to create any meaningful tests for this.

stdedos:
@Medstar117 probably the loss of coverage is because of the new functionality you've added.

Since I don't see CI running a mock test on Windows (and expect any kind of meaninful results), why don't you split the Windows stuff? It will give better visibility

Medstar117:
Oh wow XD Didn't think anyone saw this; been so long since I looked at this that I thought I was shoved to the bottom of the barrel!

Thank you for the reply @stdedos; so you're saying I should split this PR into two separate ones--one for Windows compatibility and one for the enhancements?

stdedos:
Well, I am waiting for my stuff to merge, and a lot of people open a bug report about the same things ... Sometimes I like to chime in 😄 

It is generally a good idea to merge separate "things" separately. It makes it easier for the owner when/if they decide to take a look at it

Medstar117:
Gotcha, I'll get on that as soon as I can. Thanks for the tip!

Also, sorry for the late reply, been away from the PC for a while XD

PR #409, opened 2019-06-20
Already-signed videos can also use the `sig=` attribute

stdedos:
_@coveralls is extremely chatty. Please @mention me to comment_


stdedos:
@coveralls is extremely chatty. Please @mention me to comment

PR #422, opened 2019-07-15
CLI: Windows Compatibility

Medstar117:
Known Issues Addressed:
#362
#202
#180 (a tiny bit)

Updates:
+ Added compatibility for Windows to "get_terminal_size" as to prevent cli.py from crashing when downloading a stream.
+ Added to the test_cli.py script for the newly added functionality (will most likely need revising).

**For Linux and OSX users,** the os.popen call for "stty size" has been replaced with a subprocess.check_output call. This tweak prevents stderr from displaying an invalid bash command error on Windows every time the progress bar is updated; because this puts the "display_progress_bar" function into a different process, this edit also has a side effect of reducing download time slightly (bonus!).

If either attempt at getting the terminal geometry fails, the default values are returned as follows in an attempt to prevent an error: rows: 80, columns: 24.


RONNCC:

PR #425, opened 2019-07-22
Probable fix for Error 403

aliakseis:
Probable fix for HTTP Error 403


hbmartin:
@aliakseis what video URLs were triggering the `VideoUnavailable` exception?

aliakseis:
It looks like I can't reproduce it anymore...

PR #426, opened 2019-07-25
Fix not generating the signature when `lsig` is present

CraftSpider:
This only fixes one part of the 403 error, #425 is required to fix regex not matching, and line 64 needs 'signature' replaced with 'sig' in some cases

Combining this with the other two fixes, I'm no longer getting any error 403

### DIFF
--- a/pytube/mixins.py
+++ b/pytube/mixins.py
@@ -38,7 +38,8 @@ def apply_signature(config_args, fmt, js):
         elif live_stream:
             raise LiveStreamError('Video is currently being streamed live')
 
-        if any([x in url for x in ['signature=', 'sig=']]) and 'lsig=' not in url:
+        if any([x in url for x in ['signature=', 'sig=']]) and \
+                'lsig=' not in url:
             # For certain videos, YouTube will just provide them pre-signed, in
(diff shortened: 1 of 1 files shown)


hbmartin:
@CraftSpider what URLs/video ID required this?

CraftSpider:
Thank you for your patience. Its been a long time, and I don't remember exactly. I made my two PRs as 'This fixed my issue and didn't introduce any I saw'. I know adding the prefix to sig was because there could exist an `lsig` param, and checking for `sig=` would incorrectly conclude it existed because that overlaps with `lsig=`, I will try some links and update my PR if I can find a still broken one, otherwise close them out

RONNCC:
@CraftSpider  is functionality in this PR still needed? Would be happy to assist in getting it merged

CraftSpider:
This line was fixed some time ago
https://github.com/nficano/pytube/blob/ee0deb6dc1c465927472f3c01f8644ddd909ce50/pytube/extract.py#L254
this PR is unneeded

RONNCC:
Thanks @CraftSpider 

Would love your assistance with other bugs/issues that you may run into -- definitely appreciate the PR!

PR #428, opened 2019-07-26
Update logging.py

jakedent:
Wrapped logger in error handling in case of handler exception.


hbmartin:
@jakedent what exception were you seeing? Generally `Except` statement should be as specific as possible.

jakedent:
> @jakedent what exception were you seeing? Generally `Except` statement should be as specific as possible.

Just a precaution to catch any exception raised during initialisation of pytube logger, handler and level.

RONNCC:
Hi @jakedent 

Thanks

PR #440, opened 2019-07-31
Fixed stream title parse

TermoZour:
fixed a bug when trying to download a song


hbmartin:
@TermoZour  fixed in my Python 3 only fork: https://github.com/hbmartin/pytube3

PR #460, opened 2019-08-25
Added functionality to register_on_complete_callback

michael-iuzzolino:
Enables users to pass args, kwargs to the callback for on_complete and on_progress.

Maybe there is a smarter way to do this, but I am downloading a massive number of videos from Youtube (>400k) with a large number of parallel workers. I needed a way to run post-processing script on each video once it was finished downloading. I found the callback for on_complete, but it lacked the ability to pass additional information from the current worker. 

The best way I figured was to fix the callback such that it would take arbitrary user-defined args and kwargs. I figured that others might want this functionality as well; however, please feel free to disregard this if it is a poor design.

Thanks!

Here's a toy example:
```python
import sys
import pytube

def download_in_progress(stream, chunk, file_handler, bytes_remaining, custom_string):
    sys.stdout.write(f'\r{custom_string} {bytes_remaining}')
    sys.stdout.flush()

def download_finished(stream, file_handle, custom_func):
    print("\nDownload complete!")
    print("Running post processing function...")
    custom_func()

def main(youtube_ID, post_proc_func):
    youtube_path = f'https://www.youtube.com/watch?v={youtube_ID}'
    video = pytube.YouTube(youtube_path)
    video.register_on_complete_callback(download_finished, custom_func=post_proc_func)
    video.register_on_progress_callback(download_in_progress, custom_string='Remaining: ')

    stream = video.streams.filter(progressive=True).order_by('resolution').desc().first()

    stream.download(output_path='downloads', filename='test')
```

### DIFF
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -77,8 +77,8 @@ class YouTube(object):
         # (Borg pattern).
         self.stream_monostate = {
             # user defined callback functions.
-            'on_progress': on_progress_callback,
-            'on_complete': on_complete_callback,
+            'on_progress': [on_progress_callback, None, None],
+            'on_complete': [on_complete_callback, None, None],
         }
 
         if proxies:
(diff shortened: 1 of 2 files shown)


hbmartin:
@michael-iuzzolino have you considered using a function closure? (i.e. returning a function from within a function)

RONNCC:
@michael-iuzzolino  closing issue -- feel free to reopen if this is still an issue. 
I would also suggest staging, or currying for the function to pass a function closure similar to @hbmartin 's suggestion above.

Thanks,

PR #465, opened 2019-09-07
Fixed the bug # 450

dart-neitro:
When I tried to use the "default_filename" I got the exception.
A few users had the same problem:
https://github.com/nficano/pytube/issues/450

`>>> from pytube import YouTube`
`>>> YouTube('http://youtube.com/watch?v=9bZkp7q19f0').streams.first().default_filename`
`Traceback (most recent call last):`
`  File "<input>", line 1, in <module>`
`  File ".../lib/python3.7/site-packages/pytube/streams.py", line 176, in default_filename
    title = self.player_config_args['title']`
`KeyError: 'title'`

### DIFF
--- a/pytube/streams.py
+++ b/pytube/streams.py
@@ -176,7 +176,7 @@ class Stream(object):
         player_config_args = self.player_config_args or {}
 
         if 'title' in player_config_args:
-            return player_config_args['player_config_args']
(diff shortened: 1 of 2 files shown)


nficano:
Thanks a lot @dart-neitro

PR #482, opened 2019-10-03
Fixed KeyError "title"

jonathanhanley:
Fixed the keyword error in streams.py

### DIFF
--- a/pytube/streams.py
+++ b/pytube/streams.py
@@ -173,7 +173,7 @@ class Stream(object):
         :returns:
             An os file system compatible filename.
         """
-        title = self.player_config_args['title']
+        title = self.player_config_args['player_response']['videoDetails']['title']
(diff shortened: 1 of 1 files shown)


jonathanhanley:
Fixed KeyError title

hbmartin:
@jonathanhanley can you provide an example URL demonstrating the need for this fix?

jonathanhanley:
I can't remember exactly what the issue was it was so long ago but as far as I remember it wouldn't download any video because of an error when accessing some data in a dictionary. I tested the code after the change and it worked before I committed. I hope this helps :)

jonathanhanley:
Feel free to send me a message and I might be able to help 👍

hbmartin:
@jonathanhanley fixed in my Python 3 only fork: https://github.com/hbmartin/pytube3

PR #545, opened 2020-01-04
Fixed cipher.py: Invalid regex and in query.py: order_by() None attribute case

movraxrsp:
The RE in line 45 was truncated and malformed and does not compile.

```
>>> re.compile( r'yt\.akamaized\.net/\)\s*\|\|\s*.*?\s*[cs]\s*&&\s*[adf]\.set\([^,]+\s*,\s*(?:encodeURIComponent\s*\()?\s*(?P<si$')
// traceback details removed 
re.error: missing >, unterminated name at position 108
```

### DIFF
--- a/pytube/cipher.py
+++ b/pytube/cipher.py
@@ -42,7 +42,7 @@ def get_initial_function_name(js):
         r'(?P<sig>[a-zA-Z0-9$]+)\s*=\s*function\(\s*a\s*\)\s*{\s*a\s*=\s*a\.split\(\s*""\s*\)',  # noqa: E501
         r'(["\'])signature\1\s*,\s*(?P<sig>[a-zA-Z0-9$]+)\(',
         r'\.sig\|\|(?P<sig>[a-zA-Z0-9$]+)\(',
-        r'yt\.akamaized\.net/\)\s*\|\|\s*.*?\s*[cs]\s*&&\s*[adf]\.set\([^,]+\s*,\s*(?:encodeURIComponent\s*\()?\s*(?P<si$',  # noqa: E501
+        r'yt\.akamaized\.net/\)\s*\|\|\s*.*?\s*[cs]\s*&&\s*[adf]\.set\([^,]+\s*,\s*(?:encodeURIComponent\s*\()?\s*(?P<sig>[a-zA-Z0-9$]+)\(',  # noqa: E501
         r'\b[cs]\s*&&\s*[adf]\.set\([^,]+\s*,\s*(?P<sig>[a-zA-Z0-9$]+)\(',  # noqa: E501
         r'\b[a-zA-Z0-9]+\s*&&\s*[a-zA-Z0-9]+\.set\([^,]+\s*,\s*(?P<sig>[a-zA-Z0-9$]+)\(',  # noqa: E501
         r'\bc\s*&&\s*a\.set\([^,]+\s*,\s*\([^)]*\)\s*\(\s*(?P<sig>[a-zA-Z0-9$]+)\(',  # noqa: E501
--- a/pytube/query.py
+++ b/pytube/query.py
@@ -179,10 +179,9 @@ class StreamQuery:
             def key(s):
                 return getattr(s, attribute_name)
 
-        fmt_streams = sorted(
-            fmt_streams,
-            key=key,
-        )
+        # We can sort in place instead of calling sorted()
+        fmt_streams.sort(key=key)
+
         return StreamQuery(fmt_streams)
 
     def desc(self):
--- a/pytube/query.py
+++ b/pytube/query.py
@@ -166,7 +166,7 @@ class StreamQuery:
         integer_attr_repr = {}
         for stream in self.fmt_streams:
             attr = getattr(stream, attribute_name)
-            if attr is not None:
+            if attr:
                 fmt_streams.append(stream)
                 num = ''.join(x for x in attr if x.isdigit())
                 integer_attr_repr[attr] = int(''.join(num)) if num else None
--- a/pytube/query.py
+++ b/pytube/query.py
@@ -162,13 +162,14 @@ class StreamQuery:
         :param str attribute_name:
             The name of the attribute to sort by.
         """
+        fmt_streams = []
         integer_attr_repr = {}
         for stream in self.fmt_streams:
             attr = getattr(stream, attribute_name)
-            if attr is None:
-                break
-            num = ''.join(x for x in attr if x.isdigit())
-            integer_attr_repr[attr] = int(''.join(num)) if num else None
+            if attr is not None:
+                fmt_streams.append(stream)
+                num = ''.join(x for x in attr if x.isdigit())
+                integer_attr_repr[attr] = int(''.join(num)) if num else None
 
         # if every attribute has an integer representation
         if integer_attr_repr and all(integer_attr_repr.values()):
@@ -179,7 +180,7 @@ class StreamQuery:
                 return getattr(s, attribute_name)
 
         fmt_streams = sorted(
-            self.fmt_streams,
+            fmt_streams,
             key=key,
         )
         return StreamQuery(fmt_streams)


hbmartin:
@movraxrsp can you provide a test for this?

movraxrsp:
It's a regular expression syntax error, which the interpreter indicates in what I pasted.

The diff in the commit shows what the intended correct regular expression should be.

The string in line 45 in the master branch ends with "(?P<si$')", which starts a named group but doesn't finish it.

Line 45 should end just like lines 43 and 44.

movraxrsp:
See Issue #550.

In ``query.py``, the ``order_by()`` method was broken when filtering by mp4 then ordering by resolution in the case when some of the mp4s were audio-only.

The problem was in line 168 if a stream broke out of the ``for`` loop with ``attr`` set to ``None``.

Then in line 176, ``key()`` tried to access the integer dictionary with that stream's ``None`` value as a key.

My code begins by looping through ``self.fmt_streams`` and checking each stream actually has the attribute that we are trying to sort on. It builds a list of those that do. Then orders that list.

Also, the ``for`` loop was breaking out immediately on the first stream that did not have the sort attribute. I allow the loop to continue through the entire ``self.fmt_streams`` list.

I also sort in place with ``sort()`` instead of calling ``sorted()``.

There are probably other test cases to verify.

swiftyy-mage:
Yeah you're right your way is better. It's best to give people the option to download these streams even if most cannot view them. If this gets merged, the documentation will need to be changed, it says that every stream in the stream list you are ordering by needs to have the attribute you are ordering by.

Also, would you be able to explain the first commit? I certainly understand the pattern was incomplete, and that if you were to search using that pattern individually it would not work. However, having gone over the function `regex_search` in `helpers.py`, it seems that there is no point to ever input a list to this function: it will either return a match on the first pattern in the list or it will raise a `RegexMatchError` on the first pattern in the list. Neither option allows the second pattern in the list to be searched. My apologies if that is a dumb question, I'm a bit new to this.

hbmartin:
@movraxrsp @swiftyy-mage I came up with a different approach to this problem that involved rewriting `order_by` ... https://github.com/hbmartin/pytube3/commit/3047241a1b50009f656667357ecd1f54b68bc286 ...  would be interesting to come up with more tests and compare results

movraxrsp:
Hi, @swiftyy-mage. Thanks for taking a look. (For reference, this is about the failed case in Issue #550. )

> the documentation will need to be changed, it says that every stream in the stream list you are ordering by needs to have the attribute you are ordering by.

I'm guessing that you're referring to this part that I wrote:

> "My code begins by looping through ``self.fmt_streams`` and checking each stream actually has the attribute that we are trying to sort on. It builds a list of those that do. Then orders that list."

This does not mean each stream in ``self.fmt_streams`` needs the sort attribute.

### Comparable and Non-Comparable ###

If a stream in ``self.fmt_streams`` does not have the sort attribute, then it is **non-comparable.**

In Issue #550, ``self.fmt_streams`` has non-comparable streams and comparable streams.

There are several possible ways the method could reasonably behave in this case.

1. Include non-comparables as minimums
2. Include non-comparables as a maximums
3. Exclude non-comparables
4. Let the method client decide in its call (via a new parameter)
5. Or something else.

A design decision is required. I decided to exclude.

[That's what my comment in #550 shows. The ``filter()`` call includes the audio-only stream. The ``order_by()`` call excludes it.](https://github.com/nficano/pytube/issues/550#issuecomment-575978361)

### Checking All the Streams ###

Right now, as soon as the method sees a non-comparable stream, it [breaks out of the loop in line 169](https://github.com/nficano/pytube/blob/66b33803adb1780a84c7cf75ab8968bd5f0fcffc/pytube/query.py#L169) and **ignores all the other streams.**

In the case of Issue #550, if the audio-only mp4 was the first stream in the list, then none of the video the method would return unexpected results.

It would be nice to sort the video mp4s and return them in the ``StreamQuery``. They do exist. The client is reasonable to expect a sorted list of the video streams regardless of whatever else may have been in the original list.

My code **tests _all_ the streams** and builds a list of any _comparable_ streams that may exist in ``self.fmt_streams``. Then it sorts all the comparable streams and returns the sorted list in the ``StreamQuery``.

Issue #550 is a case where the ``order_by`` **should check all the streams** _even if one of the streams is not comparable_.

movraxrsp:
Thanks, @hbmartin ... without having enough time to study your code much right now, it's obviously much more general than my ad hoc. And yes, I'd love to see how both handle a complete set of test cases. I'm jealous of the time you had to put into it.

movraxrsp:
> Also, would you be able to explain the first commit? I certainly understand the pattern was incomplete, and that if you were to search using that pattern individually it would not work. However, having gone over the function `regex_search` in `helpers.py`, it seems that there is no point to ever input a list to this function: it will either return a match on the first pattern in the list or it will raise a `RegexMatchError` on the first pattern in the list. Neither option allows the second pattern in the list to be searched. My apologies if that is a dumb question, I'm a bit new to this.

@swiftyy-mage.. interesting... I did not check ``regex_search()`` when I noticed the bad re. When I get time, I will check it out :)

(And... if any dumb questions exist, yours is not one.)

movraxrsp:
@swiftyy-mage Good catch! I agree.

I traced it. In the whole project, [line 54 in cipher.py is the only time ``regex_search()`` is called with a list argument](https://github.com/nficano/pytube/blob/66b33803adb1780a84c7cf75ab8968bd5f0fcffc/pytube/cipher.py#L54).

In its call stack, there is [only one try block](https://github.com/nficano/pytube/blob/66b33803adb1780a84c7cf75ab8968bd5f0fcffc/pytube/__main__.py#L145), and it does not catch a ``RegexMatchError``.

So if [that match ever failed, then that exception would be raised](https://github.com/nficano/pytube/blob/66b33803adb1780a84c7cf75ab8968bd5f0fcffc/pytube/helpers.py#L39) and never be handled.

I also traced the blame trail, and this error has been around since Sep 9, 2018 in [Commit c47f631](https://github.com/nficano/pytube/commit/c47f6317210409fe3b9b70a738bd8055f4faa2cd).

This really makes no sense unless I'm missing something, but I feel like I reviewed it pretty good.

swiftyy-mage:
Hello, thanks for responding. What I meant by my comment about documentation was actually just literally about the README.md where it says:

> Note that `order_by` cannot be used if your attribute is undefined in any of the Stream instances, so be sure to apply a filter to remove those before calling it.

I just meant that if this gets merged, the above statement will not be always true. I checked the actual docs and it doesn't mention it there, so they would not need to be changed.

Maybe I should make a PR with the `regex_search` function updated. Thoughts?

movraxrsp:
Thanks, makes more sense now. I do see that statement now in the README, but it is sort of buried and does not have an explicit example.

I do think ``order_by()`` should be redesigned. It is reasonable for a general client to expect that filtering for mp4s and sorting by resolution would not raise an unhandled exception, especially when there are several video mp4s in the stream list, and especially when the _very first_ example in the README is ordering mp4s by resolution without filtering twice.

(For reference, see Issue #550)

Right now, it requires reading the documentation closely and filtering twice.

//

For ``regex_search()``, go for it! I think I already fixed it in my fork, but I didn't submit a PR because I don't have any hope of a PR being seriously reviewed and committed here any time soon.

Thanks again for the discussion.

hbmartin:
@movraxrsp The implementation throws an error when sorting `int` values (like `itag`). Test cases will fail on this change. I also added more test cases in my fork which you can see here: https://github.com/hbmartin/pytube3/commit/3047241a1b50009f656667357ecd1f54b68bc286#diff-a4e953707ec9aa91a057edd0ded9244cR62
And fwiw, when testing order_by on string values with some `None`s, my implementation took 1.80s in 100,000 loops, this one took 1.86s 🤷‍♂

movraxrsp:
@hbmartin Thanks for testing it properly. I'll check everything out later when I get a min.

hbmartin:
@movraxrsp @swiftyy-mage  so cool to be talking with you about this 😄 
I added the fixed regex and updated docs (per above discussion) in my fork https://github.com/hbmartin/pytube3/commit/5f5543f24c660ab143f40b1ea67c3453bab0c9b2

movraxrsp:
### int Fail in query.py ### 
@hbmartin I just ran a test with order_by('itag'), and I see that it fails [in query.py:170](https://github.com/nficano/pytube/blob/66b33803adb1780a84c7cf75ab8968bd5f0fcffc/pytube/query.py#L170) because, of course, ints are not iterable.

This has been in the code since June of 2018... [in this commit](https://github.com/nficano/pytube/pull/288/commits/09a522b404ef28fd852d0e90b862641cb1bb7d11)

I just assumed that existing code was good and copied it up into my if statement.

🤷‍♂🤷‍♂ Lesson learned on this project.

### regex_search() ###
``regex_search()`` in [helpers.py](https://github.com/nficano/pytube/blob/66b33803adb1780a84c7cf75ab8968bd5f0fcffc/pytube/helpers.py#L38) still has problems.

In the case when a list of pattern strings is passed in and the first pattern does not match anything, an exception is raised, so none of the other patterns are checked. And... there's no code in the whole project that handles the exception if it is raised.

It looks like the code in the else block for the single pattern string case (on line 59) was just blindly copied into the list case. (Just like I did for ``order_by()``, doh!)

This comes into play in [cipher.py](https://github.com/nficano/pytube/blob/66b33803adb1780a84c7cf75ab8968bd5f0fcffc/pytube/cipher.py#L39) when ``get_initial_function_name()`` calls it with a list of patterns. I'm pretty sure this is the only call in the whole project that calls ``regex_search()`` with a list.

@swiftyy-mage knows about this too and may fix it. I'll play with it whenever I get a chance.

I'll close this PR since you've pulled fixes into your fork for these two issues. Thanks!

PR #546, opened 2020-01-05
fix KeyError for 'url'

nf-fujimoto:
Sometimes we encounted KeyError for 'url' Because of under the lines.
type of cipher_url is tuple, not dict. so it could not add url including
in stream_data['url_encoded_fmt_stream_map'].
so i changed way to add stream_data['url_encoded_fmt_stream_map'].


hrchlhck (approved):

hbmartin:
@nf-mugicha @hrchlhck fixed in my Python 3 only fork: https://github.com/hbmartin/pytube3

RONNCC:

PR #605, opened 2020-04-03
Fix typo in Playlists section in Readme

ZiggerZZ:
`video` is a string so getting `AttributeError: 'str' object has no attribute 'streams'`


RONNCC:

PR #630, opened 2020-05-18
Add logic to detect occurences of signatureCipher in stream_data formats.

jayathungek:
Occasionally, YouTube does not include the ```cipher``` key in the ```stream_data``` dict. In these cases, it uses ```signatureCipher``` instead. I've added some logic to detect this and swap the former for the latter.

In my experience, this only happens when the videos are part of a playlist, and even then only sometimes.


jayathungek:
Duplicate. Please see Issue #631

PR #631, opened 2020-05-18
add logic to detect occurences of signatureCipher in stream_data formats

jayathungek:
Occasionally, YouTube does not include the ```cipher``` key in the ```stream_data``` dict. In these cases, it uses ```signatureCipher``` instead. I've added some logic to detect this and swap the former for the latter.

In my experience, this only happens when the videos are part of a playlist, and even then only sometimes.


loveagri:
Due to limitations of youtube playlists you can only download the first 100 videos from the playlist. If you want to download more, you'll need to split your playlist into multiple seperate ones with 100 videos each. my question is how to set to download the 101-200, which function should be changed, or which params should be given.  it is better to show a example code, thanks.

RONNCC:
@jayathungek  could you add a unit test for this code and then I can merge it :)

jayathungek:
@RONNCC It seems that something has changed recently with YouTube that breaks Pytube altogether - I'm no longer able to download videos at all. I'm therefore unable to test that this functionality works as it should. It's quite late where I am, so I'll write up a more detailed issue tomorrow, but something seems to be breaking in the ```cipher.py``` module.

ritiek:
@jayathungek You mean #701?

jayathungek:
@ritiek ah yes that's the one, thanks! I've installed that patch now and everything seems to be working fine. 

@RONNCC I've added a unit test plus an extra mock to the tests directory. It passes the build tests on my end, so it should be good to go. let me know what you think :)

RONNCC:
@jayathungek  Sorry can you rebase on master -- it shows. But looks good to me besides 😄 

```
This branch has conflicts that must be resolved
```

RONNCC:
@jayathungek  merged -- would love your assistance with fixing the CI if you have time 😅

jayathungek:
@RONNCC  cheers! Just checked the Travis log and it seems to think that there aren't enough spaces before the inline comment on line 111 of pytube/cipher.py. Looking at that file, I can see that there's only 1 space there. Perhaps the format checker doesn't like this?

RONNCC:
@jayathungek  would you mind submitting a patch :)

jayathungek:
@RONNCC I've submitted a pull request - #720 that passes the CI build. Have a look at my PR message though - I'm not entirely happy with how I've accomplished this :sweat_smile:

IsmyB:
Hi, im trying to use the BOTm, but when i run it and after i select to download only in MP3, the bot download in MP4

please help, thanks

PR #638, opened 2020-05-23
Provide backward compability to Python 3.5

MichaelKlemm:
I know features of 3.7 and newer are sexy and everyone loves using them. Unfortunately there are a lot of devices supporting only version 3.5 of python. For my personal usage I made this library backward compatible, but maybe there is also interest to bring this "downgrade" also to master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,5 @@ test/**/*.xml
 .idea/tasks.xml
 .idea/modules.xml
 .idea/*.iml
+venv/
+.python-version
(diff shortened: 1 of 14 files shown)


RONNCC:
@MichaelKlemm  could you update this PR and we can merge

tfdahlin:
@RONNCC after my current PRs get merged, I'll work on making the repo work with python 3.5

RONNCC:
@tfdahlin  since Python 3.5 is EOL'd I would suggest dropping support here unfortunately.

Issue #621, opened 2020-04-27
Is there a way to get the Music-Metas from the bottom of the descr. ?


DerZweiteTyp:
Is it possible to get these values? 
http://prntscr.com/s71vvv

tooxo:
Yes, this is possible, the data is provided inside the watch_html. I will open a PR to implement this feature.

PR #683, opened 2020-07-19
Implement Copyright Notices for Music usually listed at the bottom of the description.

tooxo:
As stated in #621 I think music meta notices should be accessible by the user, so I wrote a implementation. Code review and change requests are heavily appreciated.

### DIFF
--- a/pytube/request.py
+++ b/pytube/request.py
@@ -17,7 +17,7 @@ def _execute_request(
     method: Optional[str] = None,
     headers: Optional[Dict[str, str]] = None,
 ) -> HTTPResponse:
-    base_headers = {"User-Agent": "Mozilla/5.0"}
+    base_headers = {"User-Agent": "Mozilla/5.0", "accept-language": "en-US,en"}
     if headers:
         base_headers.update(headers)
     if url.lower().startswith("http"):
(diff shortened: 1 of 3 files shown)


pytube/request.py
```diff
@@ -17,7 +17,7 @@ def _execute_request(
     method: Optional[str] = None,
     headers: Optional[Dict[str, str]] = None,
 ) -> HTTPResponse:
-    base_headers = {"User-Agent": "Mozilla/5.0"}
```
RONNCC, line 20: qq: what is rationale to add  `en-US,en` here?

tooxo: I am not from a country where english is spoken, so if I request for example a private video, the response would contain "Dieses Video ist Privat" instead of "This video is private". This fact breaks many features for non english countries like for example the last change in playlists or the upload date, because the regex-es used are made for americans. This header ensures the answer is sent in english.

RONNCC: @tooxo very good point - thank you for the context 😄

RONNCC:
@tfdahlin  can you review when you get a minute

tfdahlin:
Gonna try to look at this sometime this week hopefully

tfdahlin:
Working on this now, hopefully going to have some changes for you sometime tonight.

tfdahlin:
Superceded by https://github.com/nficano/pytube/pull/809

I was going to try to make the changes in tooxo's repo, but the branch was something like 40 commits behind the upstream master, and I didn't want to deal with resolving any merge issues. I took the opportunity to rewrite some of the logic to be cleaner as well.

RONNCC:
Thanks!

Issue #661, opened 2020-06-16
KeyError: 'url' and KeyError: 'cipher' ,please someone update this code or some procedure to fix my code


saranshbamania-iitb:
while downloading youtube video from python i am getting key error.
error:
![Screenshot (4)](https://user-images.githubusercontent.com/49087533/84765066-87ac1600-af83-11ea-925d-b1721ed648cd.png)

tristanlatr:
Having the same error on macOS Catalina

Python
```
% python3 --version
Python 3.7.6
```

Traceback
```
% pytube3 --list "https://www.youtube.com/watch?v=7tLq5FZ0xlY"
Loading video...
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/pytube/extract.py", line 297, in apply_descrambler
    for format_item in formats
  File "/usr/local/lib/python3.7/site-packages/pytube/extract.py", line 297, in <listcomp>
    for format_item in formats
KeyError: 'url'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/pytube3", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/pytube/cli.py", line 48, in main
    youtube = YouTube(args.url)
  File "/usr/local/lib/python3.7/site-packages/pytube/__main__.py", line 92, in __init__
    self.descramble()
  File "/usr/local/lib/python3.7/site-packages/pytube/__main__.py", line 132, in descramble
    apply_descrambler(self.player_config_args, fmt)
  File "/usr/local/lib/python3.7/site-packages/pytube/extract.py", line 301, in apply_descrambler
    parse_qs(formats[i]["cipher"]) for i, data in enumerate(formats)
  File "/usr/local/lib/python3.7/site-packages/pytube/extract.py", line 301, in <listcomp>
    parse_qs(formats[i]["cipher"]) for i, data in enumerate(formats)
KeyError: 'cipher'

```

tristanlatr:
@Hack-Saransh 

Here is the workaround:

kbuilds commented 18 days ago
While waiting for the maintainer to address these PRs, feel free to use our public version with #643 merged in:

https://gitlab.com/obuilds/public/pytube

I use GitLab because I am a savage
You can install the tagged version like so:

pip install git+https://gitlab.com/obuilds/public/pytube@ob-v1
or add the following to your requirements.txt

git+https://gitlab.com/obuilds/public/pytube@ob-v1
[...]

PR #685, opened 2020-07-20
Fix for issue: https://github.com/nficano/pytube/issues/661 and default video titles not being shown

isaurav0:
Address the issue: 
KeyError: 'url' and KeyError: 'cipher' ,please someone update this code or some procedure to fix my code #661
Link: https://github.com/nficano/pytube/issues/661

Cause of the issue:
Youtube has modified json response for the streams and replaced "cipher" key with "signatureCipher".

Other changes:
Youtube has begun to send titles of the video in <meta> tags. So, I have extracted titles from <meta> tags.

### DIFF
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -7,13 +7,12 @@ exclusively on the developer interface. Pytube offloads the heavy lifting to
 smaller peripheral modules and functions.
 
 """
+
 import json
 import logging
-from html import unescape
-from typing import Dict
-from typing import List
-from typing import Optional
+from typing import Optional, Dict, List
 from urllib.parse import parse_qsl
+from html import unescape
 
 from pytube import Caption
 from pytube import CaptionQuery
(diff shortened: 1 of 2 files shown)


pytube/__main__.py
```diff
@@ -120,20 +110,25 @@ def descramble(self) -> None:
...
-                    "<title>"
-                )
-                i_end = self.watch_html.lower().index("</title>")
-                title = self.watch_html[i_start:i_end].strip()
-                index = title.lower().rfind(" - youtube")
+            # ---> ADD THIS PART <---
+            if 'title' not in self.player_config_args:
+                # for more reliability when parsing, we may use a trained parser
+                try:
+                    from bs4 import BeautifulSoup
```
tooxo, line 120: I think for only extracting the title, a html parser is way too resource heavy. It is easily doable with a regex only.

PR #687, opened 2020-07-22
Adapt playlist extraction to YouTubes architectual changes

### LINKED ISSUE #670: Python playlist is returning empty array
Versions used:
pytube3 9.6.4
python 3.8.3

Python playlist is returning empty array

pl = Playlist("https://www.youtube.com/playlist?list=UUqvtaBINBx7k6pQBR5CKYzw")

### LINKED ISSUE #684: Playlist can't get more than 100 videos
Hi, 
I am trying to download a playlist with more than 100 videos, but can only get 100 urls.
I think the problem may occur from playlist.py's `def _find_load_more_url(req: str)`, may google change something so the function failed. However i am not familiar it.
Can someone give a advice?
Thx

tooxo:
This PR addresses multiple errors with playlist video extraction. 

Fixes: #684 
Fixes: #670 

and possibly other issues.

This PR is necessary, because YouTube recently changed something at their backand which changed the format of the playlist html responses, so the old way of extracting the youtube urls won't work anymore.

YouTube is now relying on putting all relevant information inside a js variable and populating the site via html. 

The corresponding js var is window["ytInitialData"] and contains json data about anything which should be displayed on the page.

The videos are now listed inside the json variable and can easily be extracted via json "tree-walking" as shown in the code. (I can add comments if neccessary, just add a change request thingie).

Another thing that changed was the way continuations/pagination works now. Before the update, getting the urls from a playlist with more than 100 songs was fairly simple, because the page contained the url which contained the next page of videos.
YouTube replaced this with a ajax api endpoint call, which needs a continuation token. This token can be extracted from the json as well as shown in the code. 
This ajax endpoint also needs some headers now (don't ask me why, I simpy found them by looking inside the developer tools and stripped the unnecessary). 
These are also included inside the code.
This endpoint contains similar data to the data initially provided inside the html page of the playlist.

This update by YouTube has a positive impact on performance of long playlists, because downloading a json is using way less time and bandwidth than downloading the entire website and extracting the urls via regex.

### DIFF
--- a/pytube/contrib/playlist.py
+++ b/pytube/contrib/playlist.py
@@ -75,14 +75,32 @@ class Playlist(Sequence):
     ) -> Iterable[List[str]]:
         """Parse the video links from the page source, yields the /watch?v=
         part from video link
+
+        :param until_watch_id Optional[str]: YouTube Video watch id until
+            which the playlist should be read.
+
+        :rtype: Iterable[List[str]]
+        :returns: Iterable of lists of YouTube watch ids
         """
         req = self.html
         videos_urls, continuation = self._extract_videos(
-            self._extract_json(req))
+            # extract the json located inside the window["ytInitialData"] js
+            # variable of the playlist html page
+            self._extract_json(req)
+        )
+        if until_watch_id:
+            try:
+                trim_index = videos_urls.index(f"/watch?v={until_watch_id}")
+                yield videos_urls[:trim_index]
+                return
+            except ValueError:
+                pass
         yield videos_urls
 
-        # The above only returns 100 or fewer links
-        # Simulating a browser request for the load more link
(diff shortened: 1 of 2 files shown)


RONNCC:
@tooxo Thank you for the PR -- could you add to the description to provide context e.g. https://medium.com/@hugooodias/the-anatomy-of-a-perfect-pull-request-567382bb6067 :)

tooxo:
Added some detail, please excuse my poor grammar I'm on mobile @RONNCC

RONNCC:
@tooxo  thanks for change! -- would love assistance with setting up auto CI testing too if have time 😉

dahmadjid:
thanks for the continued support!

tooxo:
@RONNCC important addition to this PR: #691 

(Also opened a ci related pr #690)

PR #689, opened 2020-07-24
replace pytube to pytube3

bneiluj:
Fix pip install


RONNCC:
@bneiluj Believe that is a different project 😄  -- but would love contributions to bring pytube up to speed for your usecases

PR #690, opened 2020-07-24
Fix CI builds to ensure functionality 

tooxo:
This PR intends to fix the constantly failing ci builds.

Things done: 
* Code formatting and linting with flake8
* fixing of tests with mocking
* added the use of the python coverage module to keep on updating the codecoverage on coveralls

Problems found (Should resolve before merging):
* There seems to be a problem with the on_progress_complete handler function, because the test for it won't work for some reason. I am not sure if this is caused by the test not working or by an issue in the code base. Could need some help here.


RONNCC:
🤔  let me look into the test

tooxo:
I'm pretty sure, that it is an issue with the tests, more specifically with the mocks implemented by other tests. If you run the test on its own it works flawlessly.

tooxo:
Yes, that indeed was the issue, the tests are now working flawlessly.

tooxo:
@RONNCC I think its ready for merging!

RONNCC:
@tooxo thank you for your work :) -- merged!

PR #709, opened 2020-08-02
Fix YouTube(). Downloading was broken, this is how to fix it.

REDS1736:
I simply commented one line (you could also remove it...) which would be sent as a parameter in a request. This parameter is invalid. Simply removing it solves the problem.

This is my first time ever contributing to any project; is there anything i've done wrong / unconventionally?


tfdahlin:
@RONNCC this can be closed, I actually found this myself while I was working on https://github.com/nficano/pytube/pull/775, so this is implemented

RONNCC:
Awesome thanks @tfdahlin 
fyi @REDS1736

Issue #731, opened 2020-08-26
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 164 (char 163)


amanjyotiparida840:
Most of the time i got error like this "json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 164 (char 163)" but in same  some of time it runs well ,I don't know what to do.......can anyone fix this bug.

I'm currently using 
Pytube3
Python 3.7.3
pip 20.1.1

mehmetbarispolat:
i have same issue

amanjyotiparida840:
@mehmetbarispolat, If you get any solution......then plz post....

404dcd:
Sometimes it will work and sometimes it throws that error. Seems to fail about 4/5 of the time.

jwbn:
Same here. Will occasionally work properly but a lot of the time I will get that error.

kbuilds:
I am taking a look at this one, but the issue appears to be in a ~60k character blob of json...

Matthewkm:
From a little digging YouTube pages where this problem occurs seem to have different HTML to ones that work.
As a result the get_ytplayer_config(html) function in extract.py can't get the correct information from the html resulting in a botched json config file.

This issue has been logged before : https://github.com/nficano/pytube/issues/724 and https://github.com/nficano/pytube/pull/726

I don't know if YouTube have recently started playing around with their HTML structure - seems strange only certain pages are like this.

Edit:
i can run the same code on the same yt page and sometimes it will work and others times not, so i guess short term workaround is just to run the download multiple times?

404dcd:
Could be my unreliable bandwidth but I think that they are gradually making this change to all pages, maybe some are cached. I can't find any working examples now. This is a critical issue I'm afraid

nicomda:
Same issue here.

EnDroidCN:
Same here. Following this issue.

jwbn:
Found this pull request from another issue (https://github.com/nficano/pytube/pull/726). Just edited extract.py and the issue _seems_ to be fixed for me. It's no longer giving me errors nearly every time.

popeldd:
Has anyone found any workarounds to this problem?

tfdahlin:
I'm going to close this issue;
The original post is using pytube3, not pytube, and the most recent post suggests that the issue has been fixed by a PR that recently got merged.

PR #733, opened 2020-08-27
Fix player response and basejs extraction from html

tooxo:
This fixes issues like #731 as well as extraction of specific urls with a new template on youtubes side. 

This is for now the only url I could find, which needs the url fix. 
https://www.youtube.com/watch?v=gAs9HZC9c7Y


jwbn:
Works well for me, thanks. Tried it on another video that had the same issue as [this](https://www.youtube.com/watch?v=gAs9HZC9c7Y) and it worked.

ocervell:
Works !

RONNCC:
@tooxo  is this still needed? seems like most recent  version fixes already

tooxo:
> @tooxo is this still needed? seems like most recent version fixes already

Yea the changes were committed here: https://github.com/a-czyrny/pytube/commit/1113b957543cbb07195b3e150f1217088fa084a1

RONNCC:
Thanks @tooxo !

Issue #593, opened 2020-03-18
Convert a playlist  to a zip file of videos


espoirMur:
This is a feature request!

Is it possible to download all videos from a playlist and convert it into a zip file?

Okay....

I am putting it here for the next HackForBest, I can come back to it.

swiftyy-mage:
It's definitely possible, let me know how it goes! Why don't you make a pull request [here](https://github.com/get-pytube/pytube3) and we can collaborate? We need more people working on this! 😁 [Here](https://docs.python.org/3/library/archiving.html) is some info on compression in Python, good luck.

espoirMur:
@swiftyy-mage, tomorrow I will be working on the project using this library again, I will make it.

SamsadSajid:
Hello, I have started working on this feature.

PR #750, opened 2020-10-01
Add the feature of converting a playlist to a zip file of videos

SamsadSajid:
- Add a new feature of converting a playlist to a zip file of videos.
- Feature request is for issue #593

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Finally *pytube* also includes a command-line utility, allowing you to quickly d
 
 ## Features
 - Support for Both Progressive & DASH Streams
-- Support for downloading complete playlist
+- Support for downloading complete playlist and zip the playlist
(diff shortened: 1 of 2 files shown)


tfdahlin:
Hi @SamsadSajid,
I don't think it makes sense to directly add functionality to convert a playlist to a zip file of videos, as there are many different forms of compression that users may want to use (zip, gzip, etc) for these files, and it wouldn't make sense to keep adding more options to this function.

I have added a return to `download_all` that provides a list of file paths for the downloaded videos so that you can wrap the result yourself in [this pull request](https://github.com/nficano/pytube/pull/793) so that it is possible to know where these files got saved in order to do any compression that you want manually. If you need to create a zip file of the videos, you can now do this yourself in your code, so I will be closing this PR

PR #761, opened 2020-10-22
Pytest workflow

tfdahlin:
I've created a GitHub action workflow for running pytest with the existing unit tests written for the library, and fixed a few os-specific issues within those tests. 

This workflow tests against python versions 3.5, 3.6, 3.7, and 3.8, but does not test against python 2 at all. One issue that I ran into while running these tests is that the codebase has some type hints that are unsupported by python 3.5, so the tests *do not* run successfully in 3.5. I don't know how extensive this problem is.

Python 3.5 introduced type hints in function parameters, while python 3.6 introduced [type hints for variables](https://docs.python.org/3/whatsnew/3.6.html#pep-526-syntax-for-variable-annotations). There seem to be variable type hints scattered throughout the code, and I didn't feel like tracking all of them down. If you look at [this action run](https://github.com/tfdahlin/pytube/actions/runs/321107104), the 3.5 test suite runs into its first problem [here](https://github.com/tfdahlin/pytube/blob/master/pytube/streams.py#L67). In order to fix the problem, we'd have to remove these type hints (which may or may not be worthwhile).

@RONNCC let me know what you think about this issue. We could drop support for 3.5, and thus remove it from the test matrix, or we could dig through the code and remove all of the type hints that cause 3.5 to fail. Worth noting that python 3.5 EOL was last month. I'll convert this from a draft PR to a normal PR after a decision is reached, and the necessary changes are made to this PR.


tfdahlin:
Added I005 to exclusion list in .flake8 because some files need to be imported in a certain, non-alphabetical order in order for pytube to work, so that rule cannot be applied. Fixing the single I004 error in `test_captions.py` fixed the travis-ci failures so it shouldn't fail any more on non-breaking changes. I might actually remove the github pytest workflow if the travis CI integration is going to be used, since I didn't realize travis was running the unit tests. I'll wait on a response before I do though

tfdahlin:
I'm actually going to split this into a couple of PRs, to address things individually and keep the commit history a little cleaner.

RONNCC:
sounds good Thanks!

PR #764, opened 2020-10-23
Changed platform-specific separators to platform-agnostic separators.

tfdahlin:
Attempting to run the test suites locally on a windows machine causes some issues with how path separators are evaluated. This fixes that issue by replacing references to `"/"` with references to `os.path.sep`, and uses `os.path.join()` for creating the paths.

### DIFF
--- a/tests/test_captions.py
+++ b/tests/test_captions.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import os
 from unittest import mock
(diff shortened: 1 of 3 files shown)


tfdahlin:
cc: @RONNCC

tfdahlin:
This change is now part of [PR 767](https://github.com/nficano/pytube/pull/767).

RONNCC:
ack -- thanks

Issue #759, opened 2020-10-17
Subtitles download 


jimmyxp01:
```
from pytube import YouTube

yt = YouTube(https://youtu.be/nI9BL4AB5Y0)
caption = yt.captions.get_by_language_code('en')
caption.xml_captions
'well i&#39...'
all =caption.generate_srt_captions()
print (all)

```
In this video , there is two subtitles

1.English automatically generated (code="en")

2.English (code="en")

I want to download only English caption, not English automatically generated. So can't download english. It's always download automatically generated. Help me.

ghost:
Just use the following function
```
def get_caption_by_language_name(yt, lang_name):
	for caption in yt.caption_tracks:
		if caption.name == lang_name:
			return caption
```
`get_caption_by_language_name(YouTube(url), 'English')` will return the English caption. Replace 'English' with 'English (auto-generated)' for the auto generated caption. Do a `print(YouTube(url).caption_tracks)` to see all available captions and their language names.

jimmyxp01:
Thanks it's work

PR #765, opened 2020-10-23
Update query.py

ghost:
In CaptionQuery class, added a method and a supporting data member to get captions by language name.
This solves issue #759


tfdahlin:
@skoare I implemented the fix suggested by urna in #779 in [this PR](https://github.com/nficano/pytube/pull/783), which I believe is a more effective solution. Can we close this PR?

PR #766, opened 2020-10-23
Updated Pipfile.lock to use latest library versions.

tfdahlin:
This should cause most of the dependabot PRs to close automatically, if merged. Running pytest with these dependencies installed seemed to be successful locally.


tfdahlin:
cc: @RONNCC

RONNCC:
@tfdahlin  should we be concerned by failing checks?

tfdahlin:
@RONNCC yeah, that's a good point, when I pinged you the CI build was still processing I think. I'll look into fixing that. I think we may want to drop the isort extension to flake8 altogether, honestly. There are a few imports that need to happen in a specific order to prevent import loops, and `flake8-isort` only cares about alphabetizing imports. What do you think about removing `flake8-isort` from the test build?

RONNCC:
@tfdahlin  sounds good to me -- as long as flake8 is there i'm not concerned about the rest as much

tfdahlin:
Removed the flake8-isort dependency, CI tests are passing again
cc: @RONNCC

RONNCC:
Thanks @tfdahlin

PR #773, opened 2020-10-25
Added continue-on-error to greeting action.

tfdahlin:
This action isn't something that should cause merge issues if it fails. Adding the continue-on-error attribute to the `first-interaction` action will prevent this from happening.


tfdahlin:
Closing this since it didn't affect the underlying issue with the greeting GitHub Action.

Issue #743, opened 2020-09-21
KeyError: 'streamingData'


Iftakharpy:
### I was trying to download https://www.youtube.com/watch?v=NdA778dQrXE

And this always raises KeyError: 'streamingData' error.

```python3
>>> YouTube("https://www.youtube.com/watch?v=NdA778dQrXE")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\LENOVO\AppData\Local\Programs\Python\Python38-32\lib\site-packages\pytube\__main__.py", line 92, in __init__
    self.descramble()
  File "C:\Users\LENOVO\AppData\Local\Programs\Python\Python38-32\lib\site-packages\pytube\__main__.py", line 132, in descramble
    apply_descrambler(self.player_config_args, fmt)
  File "C:\Users\LENOVO\AppData\Local\Programs\Python\Python38-32\lib\site-packages\pytube\extract.py", line 281, in apply_descrambler
    formats = json.loads(stream_data["player_response"])["streamingData"]["formats"]
KeyError: 'streamingData'
```

List of some videos raises the same error.
- https://www.youtube.com/watch?v=ltneTZ9uRGw
- https://www.youtube.com/watch?v=fp3vOzVQL5c

ghost:
It is working for me. Can you try it once again? If the problem persists, reinstall pytube from the repository. I am using the repository version.

Iftakharpy:
I am using:
Python             3.8.5
pytube             9.6.0
pytube3           9.6.4

still having the same problem.

ghost:
> I am using:
> Python 3.8.5
> pytube 9.6.0
> pytube3 9.6.4
> 
> still having the same problem.

Yeah, I also got an error when I tried it again. Sometimes it works, sometimes it doesn't. Its a pytube problem.

yehya-dev:
I am having the same issue! Not with all videos but some videos.

andipahlevy:
i am having the same issue

tfdahlin:
If you install both pytube3 and pytube with pip, certain files get overwritten in unintended ways. This is especially true because the pypi packages haven't been updated in several months.

You should be able to fix this problem if you do the following:
1. Run `pip uninstall pytube`
2. Run `pip uninstall pytube3`
3. Run `python -m pip install git+https://github.com/nficano/pytube`

This should uninstall the conflicting versions of pytube, and install the most recent code from this repository.

raghav330:
[...]

PR #775, opened 2020-10-27
Revamp of unit tests

tfdahlin:
With changes to YouTube that break pytube, it will sometimes be necessary to generate new mock files for the test suite. This adds that functionality to the `helpers.py` file, and also creates unit testing for that new function. 

As part of this pull request, the previous mock file that references Gangnam Style has been replaced with a mock file that references YouTube Rewind 2019. Additionally, existing unit tests have been updated to reflect this new mock file. 

Finally, there is another fix to make another existing unit test os-agnostic.


tfdahlin:
cc: @RONNCC

tfdahlin:
Technically this functionality *sorta* exists in the cli, but lacks unit-tests, doesn't save the result to the mock folder, and makes use of timestamps. It looks like that functionality is meant more for debugging. This simplifies the process of creating these files and includes unit test coverage for the function.

tfdahlin:
Looks like the travis CI is failing due to youtube rate limiting. I'll figure out a way to create a mock instead so we can avoid that.

tfdahlin:
This one's a bit of a doozie.
Firstly, it gets rid of the unit tests against gangnam style, in favor of unit tests against YouTube Rewind 2019, as requested as a follow-up to https://github.com/nficano/pytube/pull/767. 

In doing this, I discovered that the mock json files were generated kind of weird, and were sort of a partial serialization of a YouTube object. At first, I tried to create a function that replicated this (without realizing one existed), but while writing unit tests to make sure that it worked correctly, I ran into problems with Travis CI. Notably, I ran into `HTTP 429` responses from YouTube, so I had to work out how to use mock to get around that. As a result of this, I discovered I needed to create a cache of different webpages to test against, so I ended up reworking some of the unit tests to use pure HTML responses from a mock file, instead of using a partial serialization. Travis CI should no longer have rate limiting issues while running the unit test that I wrote for my serialization function. 

I also included another minor fix to unit tests where it was using platform-specific separators, rather than platform-agnostic.

Finally, I also attempted to follow up on #743 and I believe the way that I refactored __main__ will fix this problem.

cc: @RONNCC

tests/test_helpers.py
```diff
@@ -90,3 +95,79 @@ def test_setup_logger(logging):
...
+    # The file header includes time metadata, so *occasionally* a single
+    #  byte will be off at the very beginning. In theory, this difference
+    #  should only affect bytes 5-8 (or [4:8] because of zero-indexing),
+    #  but I've excluded the 10-byte metadata header altogether from the
+    #  check, just to be safe.
+    # Source: https://en.wikipedia.org/wiki/Gzip#File_format
+    assert gzip_data[10:] == full_content[10:]
+
+
+def test_uniqueify():
```
RONNCC, line 165: We can use python sets for this -- we dont need a uniqueify function. 

In scientific/number crunching context there's also numpy.unique

RONNCC (changes requested):

RONNCC:
@tfdahlin  one nit -- but otherwise looks good

tests/test_helpers.py
tfdahlin: @RONNCC unfortunately, we *do* need to use the existing uniqueify function. While python sets can be used for this, the problem is that converting from list -> set -> list loses the ordering. 

In the case of processing playlists (the only location uniqueify is used), the ordering of the video list is important, so this uniqueify function serves a purpose that cannot be accomplished simply by using sets.

tfdahlin:
see response in conversation

tests/test_helpers.py
```diff
@@ -90,3 +95,79 @@ def test_setup_logger(logging):
...
+
+    file_handle = mock_open.return_value.__enter__.return_value
+
+    # For some reason, write gets called multiple times, so we have to
+    #  concatenate all the write calls to get the full data before we compare
+    #  it to the BytesIO object value.
+    full_content = b''
+    for call in file_handle.write.call_args_list:
+        args, kwargs = call
+        for arg in args:
```
RONNCC: nit: less verbose? --  full_content = ''.join(args)

tests/test_helpers.py
RONNCC: good point -- can use ordered dict (python 3.5+ dicts are ordered by default also, that that's breaking API contract a bit) 

https://repl.it/@spamspam3/VoluminousSilverPrograms

tests/test_helpers.py
```diff
@@ -90,3 +95,79 @@ def test_setup_logger(logging):
...
+
+
+def test_uniqueify():
+    non_unique_list = [1, 2, 3, 3, 4, 5]
+    expected = [1, 2, 3, 4, 5]
+    result = uniqueify(non_unique_list)
+    assert len(expected) == len(result)
+
+    for i in range(len(expected)):
+        assert expected[i] == result[i]
```
RONNCC: can just use assert list1 == list2. It checks order, type, and index. 
though usually you would subclass unittest pkg for these hmm

RONNCC:
@tfdahlin  responded

tfdahlin:
Good point on those tests, changed according to your suggestions @RONNCC

RONNCC:
this is a giant set of changes -- thank you @tfdahlin  😄

PR #783, opened 2020-10-30
Improved caption selection, and updated tests to reflect the change.

### LINKED ISSUE #779: NOT able to get both "English" and "English (auto-generated)" captions by captions.get_by_language_code('en') method
In file captions.py --> class Caption:  --> __init__(), we set self.code = caption_track['languageCode']

But for "English" and "English (auto-generated)" captions, the 'languageCode' are the same as "en". So if both presented, the previous one will be lost.

Currently I just modified it by using caption_track['vssId'] instead of caption_track['languageCode'], so "English" code be ".en" and "English (auto-generated)" code by "a.en".

It works for me this way, hope this issue in pytube repository be solved.

tfdahlin:
Fixes #779 (fix provided by @urna)

### DIFF
--- a/pytube/captions.py
+++ b/pytube/captions.py
@@ -23,7 +23,12 @@ class Caption:
         """
         self.url = caption_track.get("baseUrl")
         self.name = caption_track["name"]["simpleText"]
-        self.code = caption_track["languageCode"]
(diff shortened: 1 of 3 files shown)


tfdahlin:
cc: @RONNCC

PR #791, opened 2020-11-02
Better unavailable video handling

### LINKED ISSUE #614: Error: get_ytplayer_config: could not find match for config_patterns
Installation (Python 3.6.9, pytube3-9.6.4):

```bash
pip3 uninstall -y pytube3
pip3 install git+https://github.com/nficano/pytube
python3
```

Test:

```python
from pytube import YouTube
YouTube('https://www.youtube.com/watch?v=5YceQ8YqYMc')
```

Output including the error:

```
  File ".local/lib/python3.6/site-packages/pytube/__main__.py", line 91, in __init__
    self.prefetch()
  File ".local/lib/python3.6/site-packages/pytube/__main__.py", line 183, in prefetch
    self.js_url = extract.js_url(self.watch_html)
  File ".local/lib/python3.6/site-packages/pytube/extract.py", line 143, in js_url
    base_js = get_ytplayer_config(html)["assets"]["js"]
  File ".local/lib/python3.6/site-packages/pytube/extract.py", line 202, in get_ytplayer_config
    raise RegexMatchError(caller="get_ytplayer_config", pattern="config_patterns")
pytube.exceptions.RegexMatchError: get_ytplayer_config: could not find match for config_patterns```
```

tfdahlin:
There was an issue sometimes where private videos were not detected correctly. This adds improved checking for whether a video is private, including a more clear pytube exception called VideoPrivate and additional unit tests. I believe this will help with https://github.com/nficano/pytube/issues/614.


tfdahlin:
cc: @RONNCC

tfdahlin:
Added some code for handling finished livestreams that do not have a recording available.

RONNCC:
@tfdahlin  lgtm mostly -- why the removal of  type hints though?

tfdahlin:
@ronncc it's to provide backwards compatibility in case we **do** decide to support python 2 at a future point in time. Additionally, there are varying levels of support for type hints in different minor versions of python 3. For example, while Python 3.5 supports function declaration type hints, it does not support variable type hints (a reason for [this issue](https://github.com/nficano/pytube/pull/638)).

While I think that type hints are pretty cool in theory, they aren't enforced at runtime, so they don't serve as a safeguard or anything. We can accomplish the same goal through well-defined docstrings, even if that solution isn't quite as cute as type-hinting.

tfdahlin:
@RONNCC I can technically merge my PRs myself now going forward, but for most of them I'll still ask you to look over them first if you don't mind. That way, I'll not only have an extra set of eyes on the changes, but also so that you can keep the pypi package up-to-date

RONNCC:
@tfdahlin  sounds good.
Interesting thought 3.5 allowed type hinting (https://docs.python.org/3.5/library/typing.html).

Also as general principle I plan to drop 2.7 support in near future (perhaps EoY, but need to decide on ramp plan).

tfdahlin:
@RONNCC python 3.5 *does* support type hints, but only as part of function declarations (e.g. arguments and return types). 3.6 introduced type annotations for variables (https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html)

RONNCC:
@tfdahlin  ah thanks for clarification 😄

PR #794, opened 2020-11-05
Extracting publish date

a-czyrny:
The missing attribute "publish_date" (based on the html attribute "datePublished") was added.


pytube/__main__.py
```diff
@@ -79,6 +80,7 @@ def __init__(
         self.player_config_args: Dict = {}  # inline js in the html containing
         self.player_response: Dict = {}
         # streams
+        self.publish_date: Optional[datetime] = None
```
tfdahlin: Can you remove the type hint here? One of my next goals is going to be making the codebase work in python 3.5, which doesn't support variable type hints like this.

pytube/extract.py
```diff
@@ -22,6 +24,21 @@
...
+def publish_date(watch_html: str) -> Optional[datetime]:
+    """Extract publish date
+    :param str watch_html:
+        The html contents of the watch page.
+    :rtype: str
+    :returns:
+        Publish date of the video.
+    """
+    try:
+        result = regex_search(r"(?<=itemprop=\"datePublished\" content=\")\d{4}-\d{2}-\d{2}", watch_html, group=0)
```
tfdahlin: In order for the CI to succeed, you'll need to assign that regex pattern to a variable, or add `  # noqa: E501` to the end of the line. I believe using another variable for the pattern is preferable here

pytube/__main__.py
```diff
@@ -166,6 +168,7 @@ def prefetch(self) -> None:
         self.watch_html = request.get(url=self.watch_url)
         if self.watch_html is None:
             raise VideoUnavailable(video_id=self.video_id)
+        self.publish_date = extract.publish_date(self.watch_html)
```
tfdahlin: Instead of assigning the publish_date variable, this should probably be an `@property` of the class

pytube/extract.py
```diff
@@ -22,6 +24,21 @@
...
+    except RegexMatchError:
```
tfdahlin, line 39: I'm not sure if we should be catching a RegexMatchError here. Are there any examples of videos without publish dates?

tfdahlin (changes requested):

pytube/extract.py
a-czyrny: In the past it happened that YouTube did not answer the requests correctly and the HTML was incomplete. This rare case I would like to catch here,

pytube/__main__.py
a-czyrny: Good point. Because not all attributes were implemented as "property" I implemented it like this. Will be changed.

pytube/extract.py
a-czyrny: I prefer the "# noqa: E501" solution. Will be added.

pytube/__main__.py
a-czyrny: Python 3.6 is almost 4 years old. Does it make sense to support 3.5? What would be a case where this makes sense.

tfdahlin: Python 3.5 got its most recent (and final) update last month, and commands somewhere around 12% market share of python installs from what I can gather. I really do think that we should encourage users to upgrade where they can, but we can't ignore the fact that python 3.5 was in active development up until just last month, even if it is an older version. We also know that a part of the pytube userbase is using python 3.5 without the ability to upgrade, because we have a [PR](https://github.com/nficano/pytube/pull/638) saying as much

In the future, I agree that we should drop support, but providing this backwards compatibility right now doesn't really have a negative impact on us.

a-czyrny: Good explanation. I adapt my changes accordingly.

tfdahlin (approved):
Once the CI stuff passes, I'll merge this since it looks good to me

PR #799, opened 2020-11-07
Fix/404 error on adaptive

### LINKED ISSUE #717: 404 on adaptive downloads
I'm getting a 404 from youtube when I attempt to download adaptive streams. Progressive is fine, but those aren't the highest quality videos. Is there a work around for this one?

```
from pytube import YouTube
yt_video = YouTube("https://www.youtube.com/watch?v=xFSVoVOvaew")
yt_video.streams.filter(adaptive=True).order_by('resolution').first().download()
```

Returns
```
  File "/Users/kaiser/tmp/pytube/pytube/streams.py", line 245, in download
    bytes_remaining = self.filesize
  File "/Users/kaiser/tmp/pytube/pytube/streams.py", line 156, in filesize
    self._filesize = request.filesize(self.url)
  File "/Users/kaiser/tmp/pytube/pytube/request.py", line 86, in filesize
    return int(head(url)["content-length"])
  File "/Users/kaiser/tmp/pytube/pytube/request.py", line 98, in head
    response_headers = _execute_request(url, method="HEAD").info()
  File "/Users/kaiser/tmp/pytube/pytube/request.py", line 27, in _execute_request
    return urlopen(request)  # nosec
  File "/Users/kaiser/.pyenv/versions/3.8.2/lib/python3.8/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/Users/kaiser/.pyenv/versions/3.8.2/lib/python3.8/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/Users/kaiser/.pyenv/versions/3.8.2/lib/python3.8/urllib/request.py", line 640, in http_response
    response = self.parent.error(
  File "/Users/kaiser/.pyenv/versions/3.8.2/lib/python3.8/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/Users/kaiser/.pyenv/versions/3.8.2/lib/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/Users/kaiser/.pyenv/versions/3.8.2/lib/python3.8/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found
```

tfdahlin:
Attempt at fixing https://github.com/nficano/pytube/issues/717
It looks like YouTube has changed how adaptive downloads work, and you have to request a sequence of files using a parameter, rather than simply being able to directly download the stream. This catches the 404 error while attempting to download the stream, and handles creating the sequential requests.

It is worth noting that YouTube seems to be stripping video metadata from these streams. This *might* be something that can be fixed at a later point in time, but diving into the raw binary of dash files to repair metadata will likely be a heavy lift.

Also fixes a minor implementation mistake in cipher.py -- the implementation of splice is incorrect, but I don't think this has caused issues in the past because it's always called with identical arguments.


tfdahlin:
@RONNCC would appreciate your eyes on this one when you have the time

pytube/cipher.py
```diff
@@ -247,7 +247,7 @@ def splice(arr: List, b: int):
     >>> splice([1, 2, 3, 4], 2)
     [1, 2]
     """
-    return arr[:b] + arr[b * 2 :]
+    return arr[:b]
```
RONNCC, line 250: can we remove splice function - seems perhaps porting of another language construct for limited marginal utility

tfdahlin: I don't think we *can* remove the splice function. Pytube basically emulates parts of the javascript in order to generate the video signature.

pytube/request.py
```diff
@@ -43,6 +48,50 @@ def get(url, extra_headers=None) -> str:
     return _execute_request(url, headers=extra_headers).read().decode("utf-8")
 
 
+def seq_stream(url, chunk_size=4096, range_size=9437184):
```
RONNCC: we should use magic numbers from  library-level constant or SoT constant if possible

pytube/request.py
```diff
@@ -86,6 +135,56 @@ def filesize(url: str) -> int:
     return int(head(url)["content-length"])
 
 
+@lru_cache(maxsize=None)
```
RONNCC: definitely should bound here, especially if this is used on low-memory devices (e.g. raspberry pi's)

pytube/request.py
tfdahlin: fine by me, I can move some of these defaults to a constants file somewhere. I basically just stole the function declaration from `stream()`, hence those values

pytube/request.py
tfdahlin: I can definitely change this to not have a maxsize argument (default value is 128), but I think it makes more sense to just cache it. Storing an integer in cache is probably going to be preferable to making a network request, especially because of YouTube's rate-limiting

pytube/request.py
RONNCC: Can just move to top of this file or class level constant probably since this is only major networking component

pytube/cipher.py
RONNCC: ah good point

pytube/request.py
RONNCC: 😄  tends to be good practice to set _some_ sensible limit to ensure one library does not unduly use resources.

pytube/request.py
```diff
@@ -86,6 +135,56 @@ def filesize(url: str) -> int:
...
+    total_filesize += len(response_value)
+
+    # We can then parse the header to find the number of segments
+    segment_count = 0
+    stream_info = response_value.split(b'\r\n')
+    segment_regex = b'Segment-Count: (\\d+)'
+    for line in stream_info:
+        try:
+            segment_count = int(regex_search(segment_regex, line, 1))
+        except RegexMatchError:
```
RONNCC, line 164: generally comment when we except an error to explain why we ignore it

RONNCC (commented):
minor nits

pytube/request.py
tfdahlin: Moved to top of file in most recent commits

pytube/request.py
tfdahlin: I removed the maxsize argument, so now it will only cache up to 128

pytube/request.py
tfdahlin: Added explanation

tfdahlin:
updated cc: @RONNCC

RONNCC:
@tfdahlin  last nit - can you add unit test for splice, it's my bad for not mentioning it earlier.
Besides that LGTM just need to condense the commit message in the squash

tfdahlin:
Yeah, I'll add the unit test for splice in a little bit

tfdahlin:
Not sure why Travis CI isn't running, but the test and flake8 passes locally.

PR #801, opened 2020-11-08
Removed typing_extensions requirement, and updated Pipfiles to match.

tfdahlin:
The README has stated that pytube is dependency-free for quite a while, but this isn't entirely true -- the `typing_extensions` library [was added](https://github.com/nficano/pytube/commit/df469edaf1bc13a43ebec72e0fc352a0beea608b#diff-230078d672f10d17463a8a6265cad825b790885898256a3365be90685caac58d) while hbmartin was developing pytube 3. 

This removes that depenency from pytube, which was only relevant for two type annotations in `monostate.py`, so pytube can truly be dependency-free.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,6 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-typing_extensions = "*"
 
 [dev-packages]
 black = "==19.10b0"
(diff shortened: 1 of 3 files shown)


tfdahlin:
cc: @RONNCC

PR #803, opened 2020-11-08
Add python version to README

RONNCC:
Adds requirement of Python 3.6+


RONNCC:
can you review  @tfdahlin

tfdahlin:
@RONNCC I don't think we really need this, since it's covered by the badge at the top of the README
image

RONNCC:
@tfdahlin  ah good point

PR #808, opened 2020-11-09
Added new config response and optimized regex

### LINKED ISSUE #806: Regex simplification (setconfig_patterns) #enhancement
https://github.com/nficano/pytube/blob/26a50cc218a37e7ee6e3fbfb1eb301b78a61bbd7/pytube/extract.py#L265

Two almost identical regex strings for different characters can easily be merged into one.

`
yt\.setConfig\((.*['\"]PLAYER_CONFIG['\"]:\s*{.+?})\);
`

### LINKED ISSUE #807: Differing config dictionary (responseContext)
Sometimes none of the current patterns match. When i checked the watch_html there was a new config dictionary with different structure and property names. 

With the new pattern `ytInitialPlayerResponse\s+?=\s+?({.+?});` you get the following dictionary: https://gist.github.com/a-czyrny/d73ca8ea7df36a11712c2f0bcb18dbea . This behavior is not determenistic and can occur randomly for certain videos. Due to the structure of the dictionary and the different properties, there is a development effort that should be discussed here.

### LINKED ISSUE #818: RegexMatchError: get_ytplayer_config: could not find match for config_patterns, setconfig_patterns
I seem to be getting the error in the title randomly as I try to iterate through this playlist and download the audio. It has happened for multiple URLs in the playlist, but not consistently. For instance, first it failed on this URL (the 5th in the playlist):

https://www.youtube.com/watch?v=-zrR2Ev_xFg

Then on the next run, it worked fine on that URL, but failed on this URL (the 12th in the playlist):

https://www.youtube.com/watch?v=lx2aL9rLzyw

Code here:

    from pytube import YouTube, Playlist

    f_path = 'D:/mashups/'
    playlist_url = 'https://www.youtube.com/playlist?list=PL1HgMl7SM77uK8CualHzOWBFC8ouXjAme'
    playlist = Playlist(playlist_url).video_urls
    for j, url in enumerate(playlist):
        stream = YouTube(url).streams.get_audio_only()
        stream.download(f_path, str(j) + ' - ' + stream.default_filename)

a-czyrny:
Added new config response and optimized regex. See https://github.com/nficano/pytube/issues/806 & https://github.com/nficano/pytube/issues/807

### DIFF
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -150,9 +150,12 @@ class YouTube:
             self.initialize_stream_objects(fmt)
 
         # load the player_response object (contains subtitle information)
-        self.player_response = json.loads(
-            self.player_config_args["player_response"]
-        )
+        if isinstance(self.player_config_args["player_response"], str):
+            self.player_response = json.loads(
+                self.player_config_args["player_response"]
+            )
+        else:
+            self.player_response = self.player_config_args["player_response"]
         del self.player_config_args["player_response"]
         self.stream_monostate.title = self.title
         self.stream_monostate.duration = self.length
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -123,8 +123,11 @@ class YouTube:
         if 'streamingData' not in self.player_config_args['player_response']:
             config_response = get_ytplayer_config(
                 self.watch_html
-            )['args']['player_response']
-            self.player_config_args['player_response'] = config_response
+            )
+            if 'args' in config_response:
+                self.player_config_args['player_response'] = config_response['args']['player_response']
+            else:
+                self.player_config_args['player_response'] = config_response
 
         # https://github.com/nficano/pytube/issues/165
         stream_maps = ["url_encoded_fmt_stream_map"]
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -121,9 +121,7 @@ class YouTube:
         # On pre-signed videos, we need to use get_ytplayer_config to fix
         #  the player_response item
         if 'streamingData' not in self.player_config_args['player_response']:
-            config_response = get_ytplayer_config(
-                self.watch_html
-            )
+            config_response = get_ytplayer_config(self.watch_html)
             if 'args' in config_response:
                 self.player_config_args['player_response'] = config_response['args']['player_response']  # noqa: E501
             else:
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -125,7 +125,7 @@ class YouTube:
                 self.watch_html
             )
             if 'args' in config_response:
-                self.player_config_args['player_response'] = config_response['args']['player_response']
+                self.player_config_args['player_response'] = config_response['args']['player_response']  # noqa: E501
             else:
                 self.player_config_args['player_response'] = config_response
 
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 """This module contains all non-cipher related data extraction logic."""
-import ast
 import json
 import logging
 import re
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -248,8 +248,8 @@ def get_ytplayer_config(html: str) -> Any:
     """
     logger.debug("finding initial function name")
     config_patterns = [
-        r";ytplayer\.config\s*=\s*({.*?(?<!gdpr)});",
-        r"ytInitialPlayerResponse\s+?=\s+?({.+?});"
+        r"ytplayer\.config\s+?=\s+?({.+?(?<!gdpr)});",
+        r"ytInitialPlayerResponse\s+?=\s+?({.+?(?<!gdpr)});"
     ]
     for pattern in config_patterns:
         regex = re.compile(pattern)
@@ -350,7 +350,10 @@ def apply_descrambler(stream_data: Dict, key: str) -> None:
     if key == "url_encoded_fmt_stream_map" and not stream_data.get(
         "url_encoded_fmt_stream_map"
     ):
-        streaming_data = json.loads(stream_data["player_response"])["streamingData"]
+        if isinstance(stream_data["player_response"], str):
+            streaming_data = json.loads(stream_data["player_response"])["streamingData"]
+        else:
+            streaming_data = stream_data["player_response"]
         formats = []
         if 'formats' in streaming_data.keys():
             formats.extend(streaming_data['formats'])
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -248,7 +248,7 @@ def get_ytplayer_config(html: str) -> Any:
     """
     logger.debug("finding initial function name")
     config_patterns = [
-        r"ytplayer\.config\s*=\s*({.+?});ytplayer",
+        r"ytplayer\.config\s*=\s*({.+?(?<!gdpr)});",
         r"ytInitialPlayerResponse\s*=\s*({.+?(?<!gdpr)});"
     ]
     for pattern in config_patterns:
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -176,7 +176,7 @@ def js_url(html: str) -> str:
     """
     try:
         base_js = get_ytplayer_config(html)['assets']['js']
-    except KeyError:
+    except (KeyError, RegexMatchError):
         base_js = get_ytplayer_js(html)
     return "https://youtube.com" + base_js
 
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -248,8 +248,8 @@ def get_ytplayer_config(html: str) -> Any:
     """
     logger.debug("finding initial function name")
     config_patterns = [
-        r"ytplayer\.config\s+?=\s+?({.+?(?<!gdpr)});",
-        r"ytInitialPlayerResponse\s+?=\s+?({.+?(?<!gdpr)});"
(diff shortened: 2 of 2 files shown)


a-czyrny:
Tested with 32 videos (2 from each year since 2005).

pytube/extract.py
```diff
@@ -263,8 +264,7 @@ def get_ytplayer_config(html: str) -> Any:
     #  and use then load that as json to find PLAYER_CONFIG
     #  inside of it.
     setconfig_patterns = [
-        r"yt\.setConfig\((.*'PLAYER_CONFIG':\s*{.+?})\);",
-        r"yt\.setConfig\((.*\"PLAYER_CONFIG\":\s*{.+?})\);"
+        r"yt\.setConfig\((.*['\"]PLAYER_CONFIG['\"]:\s*{.+?})\);"
```
tfdahlin, line 267: Good catch. At one point while I was experimenting with these regexes I had this, not sure what happened to that.

pytube/extract.py
```diff
@@ -248,7 +248,8 @@ def get_ytplayer_config(html: str) -> Any:
     """
     logger.debug("finding initial function name")
     config_patterns = [
-        r";ytplayer\.config\s*=\s*({.*?});",
+        r";ytplayer\.config\s*=\s*({.*?(?<!gdpr)});",
```
tfdahlin: What's the context for the gdpr bit? Do you happen to have an example video where we need to capture this?

pytube/__main__.py
```diff
@@ -123,8 +123,11 @@ def descramble(self) -> None:
         if 'streamingData' not in self.player_config_args['player_response']:
             config_response = get_ytplayer_config(
```
tfdahlin: lines 124-126 can be compressed to a single line, `config_response = get_ytplayer_config(self.watch_html)` now that the logic for setting `self.player_config_args` has been split off of it

tfdahlin (changes requested):

pytube/extract.py
a-czyrny: Unfortunately it happens rarely and not reliably that the "gdpr" appears in the dictionary. It is then enclosed by two curly brackets with a following semicolon. This would lead to a premature hit and the dictionary would be invalid. Therefore this string is ignored.

pytube/__main__.py
a-czyrny: Indeed.

pytube/extract.py
tfdahlin: I'll take your word for it then.

tfdahlin:
I'll merge this a little later today, just want to run it through a few test videos myself before I do.

a-czyrny:
> I'll merge this a little later today, just want to run it through a few test videos myself before I do.

All right. Wouldn't do otherwise.

tfdahlin:
Funnily enough, I'm getting an issue when I try `YouTube('https://www.youtube.com/watch?v=9bZkp7q19f0').streams.first.download()` occasionally. Going to try and figure out what's happening.

tfdahlin:
Oh, wait, I may not have installed your branch correctly. Let me try to fix that first.

tfdahlin:
No, still running into the issue. Gonna see if I can figure out why.

a-czyrny:
The deviating dictionary requires a close examination. I still get in rare cases an exception because it cannot be parsed correctly.

tfdahlin:
Let's see if we can resolve the issue before merging this then, I'll be able to dedicate more time to it this afternoon

tfdahlin:
Sorry, gonna have to do this tomorrow, some stuff came up, and I got distracted by another PR i was working on

a-czyrny:
Here is an example for a watch page with the "{gdpr};" string inside the ytplayer.config dictionary:

https://gist.github.com/a-czyrny/3406242315c2503a5971c667bed8c816

tfdahlin:
I think I found a better way to bound the end of the string we're looking for, it works successfully against the videos I've been testing it against locally, and still passes all of the existing unit tests. Can you see if you have any issues with it @a-czyrny? I removed the extra gdpr bit in one of the regexes because I don't think it will be relevant any more with this new regex.

a-czyrny:
@tfdahlin Your solution does not have the flexibility of my solution. If the string does not end with " ytplayer" there is no match although the dictionary is present (see the new ytInitialPlayerResponse for example). In my solution it doesn't matter what comes after the semicolon.  If necessary the special feature of "gdpr" will be handled as well.

tfdahlin:
@a-czyrny i was running into issues with your regex, notably with the gangnam style video, where the json wouldn't decode properly. When I looked at the the raw HTML for pages that didn't parse correctly, I was consistently seeing the json dictionary end, and then another ytplayer element get set:
image
This seemed to be the case with all of the videos that I was testing against, and this change also passed all of the unit tests successfully without needing to make changes. The string that we're looking for, where `ytplayer.config` gets set, is part of a larger <script> tag that also sets `ytplayer.web_player_context_config` and `ytplayer.load`.

a-czyrny:
The feedback on my adjustments is very positive (https://github.com/nficano/pytube/issues/815#issuecomment-726350383, https://github.com/nficano/pytube/issues/815#issuecomment-726432525) so I see no reason to change them. Otherwise I need the URL that leads to an error.

tfdahlin:
Both of those comments came *after* I made the changes (I pushed my change 3 days ago, those comments were yesterday) :p they're using the version with my regex in it

a-czyrny:
I missed that. So speaks for you. Nevertheless I would like to see an example where my regex pattern breaks. I still see the danger that your solution with the fixed extension/order in the string will become obsolete faster if only a small change is made (e.g. the order in the script, another following key etc.).

tfdahlin:
For me, it regularly broke on this video: https://www.youtube.com/watch?v=9bZkp7q19f0
It was a little inconsistent, but after a try or two i ran into the issue

tfdahlin:
```python
from pytube import YouTube, Playlist
from urllib import parse
import sys, traceback

test_ids = [
    'DkffTDkSADI',  # Indochine - Song for a Dream (Clip officiel)
    'XhFzKhTFY3I',  # Indochine - La vie est belle (Clip officiel)
    'fJ9rUzIMcZQ',  # Queen - Bohemian Rhapsody
    '9bZkp7q19f0',  # PSY - GANGNAM STYLE(강남스타일) M/V
    'UGN00k2bHgs',  # Hustle (P!nk)
    'Zi_XLOBDo_Y',  # Billie Jean
    'dQw4w9WgXcQ',  # Never Gonna Give You Up
]

bad_urls = []
exceptions = []
tracebacks = []

for id in test_ids:
    try:
        url = f'https://www.youtube.com/watch?v={id}'
        yt = YouTube(url)
        #yt.streams.first().download(skip_existing=False)
    except Exception as e:
        #with open(f'bad_html_{id}.html', 'wb') as f:
        #    f.write(yt.watch_html.encode('utf-8'))
        exceptions.append(e)
        tracebacks.append(traceback.format_exc())
        bad_urls.append(url)

for i in range(len(exceptions)):
    print(bad_urls[i])
    print(exceptions[i])
    print(tracebacks[i]) 
```
This is the code I was using for testing, I think I also had problems with the billie jean video

a-czyrny:
Thanks for the example, I will debug with it later.

a-czyrny:
I found a new regex pattern `>\s*?yt.setConfig\(({.+?})\);` that's leads to: https://gist.github.com/a-czyrny/7dc53b22189c7f958b50d462829f7d1c. For example the key length_seconds/lengthSeconds is changed to videoDurationSeconds. None of my patterns fail no matter how often I run it, only the new pattern (untreated) fails.

tfdahlin:
was this on an age-restricted video by chance?

a-czyrny:
Matches for "Indochine - La vie est belle (Clip officiel)".

tfdahlin:
Yeah, I suspected that may be an age-restricted video. It *may* be worth refactoring some of the code to extract different things based on whether the video is age restricted or not. It sounds like the embed html might differ enough to warrant that

a-czyrny:
I agree. There are at least 3 dictionaries of which one has different keys.

a-czyrny (commented):
Can we agree that your changes are not "necessary" and we leave it at my variant. The missing patterns have to be implemented separately as discussed and throw an exception for the time being.

tfdahlin:
Yeah, lets revert my changes, merge this, and work on a new PR for the different variants

tfdahlin:
@a-czyrny I reverted my change, but the test suite I was using failed on three of the videos:
https://www.youtube.com/watch?v=DkffTDkSADI
https://www.youtube.com/watch?v=XhFzKhTFY3I
https://www.youtube.com/watch?v=Zi_XLOBDo_Y

I found out that there were a couple similar cases to what you found with the `{gdpr};` example, but with different values:
`{GDPR};`
`{ivc_transaction}`
Because these are different lengths, we can't simply change your look-behind to something like `(?<!(gdpr)|(GDPR)|(ivc_transaction))`, as look-behinds require fixed-width patterns.
However, I don't run into these problems at all with the addition of `;ytplayer` to the end of the string like I do with this reversion. I think that for the time being, my solution of adding `;ytplayer` to the end is more effective given the current state of youtube.

I tried looking into things a little bit more to see if we could potentially create a more robust regex for this, but then I realized that a simple regex can't solve this problem -- we'll need a context-aware parser for js. That's a very heavy lift though, and should be a stretch goal.

a-czyrny:
Then keep your last change. Thank you for the effort.

PR #809, opened 2020-11-10
[Feature] Video metadata

tfdahlin:
Based off of https://github.com/nficano/pytube/pull/683, but more generalized.

Includes a fix for users whose default language is not English, where certain regexes wouldn't work.
This accesses the metadata associated with the video, and makes it accessible like a dict.

### DIFF
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -442,5 +442,4 @@ def metadata(initial_data) -> Optional[YouTubeMetadata]:
     #  and build a metadata object from this new list
     metadata_rows = [x["metadataRowRenderer"] for x in metadata_rows]
 
-    ytmd = YouTubeMetadata(metadata_rows)
-    return ytmd
+    return YouTubeMetadata(metadata_rows)
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -426,7 +426,7 @@ def metadata(initial_data) -> Optional[YouTubeMetadata]:
         metadata_rows: List = initial_data["contents"]["twoColumnWatchNextResults"][
             "results"]["results"]["contents"][1]["videoSecondaryInfoRenderer"][
             "metadataRowContainer"]["metadataRowContainerRenderer"]["rows"]
-    except (KeyError, IndexError) as e:
+    except (KeyError, IndexError):
         # If there's an exception accessing this data, it probably doesn't exist.
         return None
 
@@ -438,7 +438,8 @@ def metadata(initial_data) -> Optional[YouTubeMetadata]:
             metadata_rows
         )
     )
-    # We then access the metadataRowRenderer key in each element, and build a license from that list
+    # We then access the metadataRowRenderer key in each element
+    #  and build a metadata object from this new list
     metadata_rows = [x["metadataRowRenderer"] for x in metadata_rows]
 
     ytmd = YouTubeMetadata(metadata_rows)
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -428,7 +428,7 @@ def metadata(initial_data) -> Optional[YouTubeMetadata]:
             "metadataRowContainer"]["metadataRowContainerRenderer"]["rows"]
     except (KeyError, IndexError):
         # If there's an exception accessing this data, it probably doesn't exist.
-        return None
+        return YouTubeMetadata([])
 
     # Rows appear to only have "metadataRowRenderer" or "metadataRowHeaderRenderer"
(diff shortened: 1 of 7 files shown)


tfdahlin:
cc: @RONNCC 

I'll be writing unit tests for this later, but would appreciate your eyes. This is a revamp of https://github.com/nficano/pytube/pull/683

tfdahlin:
@RONNCC this is ready now, please look over it when you're able to.

pytube/extract.py
```diff
@@ -396,3 +398,48 @@ def apply_descrambler(stream_data: Dict, key: str) -> None:
...
+    try:
+        match = regex_search(initial_data_pattern, watch_html, 1)
+    except RegexMatchError:
+        return "{}"
+    else:
+        return match[:-1]
+
+
+def metadata(initial_data) -> Optional[YouTubeMetadata]:
+    """Get the metadata for the video.
```
RONNCC: adding example of some metadata would be helpful here when looking back at code :)

pytube/extract.py
```diff
@@ -396,3 +398,48 @@ def apply_descrambler(stream_data: Dict, key: str) -> None:
         ]
 
     logger.debug("applying descrambler")
+
+
+def initial_data(watch_html: str) -> str:
+    """
+    Extracts the initial_data json from the watch_html page.
```
RONNCC: Extracts the initial_data json from the watch_html page. ->     Extracts a json from youtube watch_html page called "initial_data" that contains metadata about the video used by other functions such as copyright information.

pytube/extract.py
```diff
@@ -396,3 +398,48 @@ def apply_descrambler(stream_data: Dict, key: str) -> None:
...
+        metadata_rows: List = initial_data["contents"]["twoColumnWatchNextResults"][
+            "results"]["results"]["contents"][1]["videoSecondaryInfoRenderer"][
+            "metadataRowContainer"]["metadataRowContainerRenderer"]["rows"]
+    except (KeyError, IndexError):
+        # If there's an exception accessing this data, it probably doesn't exist.
+        return YouTubeMetadata([])
+
+    # Rows appear to only have "metadataRowRenderer" or "metadataRowHeaderRenderer"
+    #  and we only care about the former, so we filter the others
+    metadata_rows = list(
```
RONNCC: nit: can combine with 443

pytube/extract.py
```diff
@@ -396,3 +398,48 @@ def apply_descrambler(stream_data: Dict, key: str) -> None:
...
+        filter(
+            lambda x: "metadataRowRenderer" in x.keys(),
+            metadata_rows
+        )
+    )
+    # We then access the metadataRowRenderer key in each element
+    #  and build a metadata object from this new list
+    metadata_rows = [x["metadataRowRenderer"] for x in metadata_rows]
```
RONNCC, line 453: [x["metadataRowRenderer"] for x in metadata_rows if "metadataRowRenderer" in x.keys()]

prbably should dump "metadataRowRenderer" somewhere in top of class and explain how we got it / if it's just youtube magic constant

pytube/metadata.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+"""This module contains the YouTubeMetadata class."""
+import json
+from typing import Dict
+from typing import List
+from typing import Optional
+
+
+class YouTubeMetadata:
+    def __init__(self, metadata: List):
```
RONNCC, line 10: instead of list we can take iterator so we don't have to convert around that much, but isnt particularly important

tfdahlin: I don't think we should take an iterator here, because the code for building the object expects a list. The metadata itself is a list of json objects, all that I'm doing in `extract.metadata` is filtering some of those elements out:
image
This __init__ implementation strictly will not work with certain types of iterators, so I believe the argument should be as specific as possible, in this case a List.

pytube/extract.py
tfdahlin: I actually explain this on 444-445. The metadata_rows are all objects with a single key (either "metadataRowRenderer" or "metadataRowHeaderRenderer"), and the value of that key is the actual metadata.

pytube/extract.py
tfdahlin: I don't think I want to combine these for the sake of brevity, however I *can* actually get rid of the list conversion here, because of my list comprehension on 453. I really don't believe that combining list operations for the sake of "less code" is good practice, because it makes it significantly harder to read and explain

PR #819, opened 2020-11-13
Playlist update

tfdahlin:
It's bothered me for a while now that about half of the code in `playlist.py` had deprecation decorators on it. This PR removes deprecated functions in playlist, converts the `Playlist.title()` function to class property `Playlist.title`, and adds some documentation for how to use the Playlist objects to the /docs directory. 

This introduces changes that are *not* backwards compatible, because certain functions of the `Playlist` class have been deleted, and the `title()` function has been replaced with the `title` property. As such, the major version should be bumped to 10.0.0, which I have included as part of this PR.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 9.7.2
+current_version = 10.0.0
 commit = True
 tag = True
 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 10.0.0
+current_version = 9.7.2
 commit = True
 tag = True
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        python: [3.5, 3.6, 3.7, 3.8]
+        python: [3.6, 3.7, 3.8, 3.9]
 
     steps:
       - name: Checkout repo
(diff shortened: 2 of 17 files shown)


tfdahlin:
cc: @RONNCC

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 9.7.2
```
RONNCC: this is auto-controlled by deploy scripts;

RONNCC: lgtm can we remove this one though

tfdahlin: Yeah, I'll revert that change. Just make sure that the deploy scripts do a major version bump, since this does introduce breaking changes

tfdahlin: Also, I'm realizing theres a bit more that I want to rework in the playlist library to bring it more in line with the rest of the codebase, I'll ping you again when it's ready.

tfdahlin:
Ideally, I would like to see #808 get merged, that update get pushed to pypi as a patch version update, and then have this get merged and pushed to pypi

RONNCC:
@tfdahlin  ack - waiting on #808

tfdahlin:
@RONNCC when you get the chance, please do a patch version bump to pypi since #808 got merged, then merge this and do a major version bump to pypi. That way folks don't have to do major version upgrade, but can still get the patch

RONNCC:
@tfdahlin  Done - 
9.7.3 and 10.0.0 respectively

RONNCC:
https://pypi.org/project/pytube/10.0.0/

PR #825, opened 2020-11-13
Feature/video keywords

### LINKED ISSUE #822: Is it possible to obtain video keywords/tags?
Is it possible to obtain video keywords/tags with PyTube? If so, how do you do it? These are usually not visible to the user directly on a YouTube video, but I know that there are certain Chrome plugins such as TubeBuddy which are capable of displaying any arbitrary video's tags/keywords.

Is it possible to obtain this information with PyTube?

tfdahlin:
Feature add that fetches video keywords.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -34,3 +34,23 @@ def test_video_unavailable(get):
     )
     with pytest.raises(VideoUnavailable):
         youtube.prefetch()
+
+
+def test_video_keywords(cipher_signature):
+    expected = [
+        'Rewind', 'Rewind 2019',
+        'youtube rewind 2019', '#YouTubeRewind',
+        'MrBeast', 'PewDiePie', 'James Charles',
+        'Shane Dawson', 'CaseyNeistat', 'RiceGum',
+        'Simone Giertz', 'JennaMarbles', 'Lilly Singh',
+        'emma chamberlain', 'The Try Guys', 'Fortnite',
+        'Minecraft', 'Roblox', 'Marshmello',
+        'Garena Free Fire', 'GTA V', 'Lachlan',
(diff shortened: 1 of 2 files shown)


tfdahlin:
cc: @RONNCC

tests/test_main.py
```diff
@@ -34,3 +34,23 @@ def test_video_unavailable(get):
...
+        'Minecraft', 'Roblox', 'Marshmello',
+        'Garena Free Fire', 'GTA V', 'Lachlan',
+        'Anaysa', 'jeffreestar', 'Noah Schnapp',
+        'Jennelle Eliana', 'T-Series', 'Azzyland',
+        'LazarBeam', 'Dude Perfect', 'David Dobrik',
+        'KSI', 'NikkieTutorials', 'Kurzgesagt',
+        'Jelly', 'Ariana Grande', 'Billie Eilish',
+        'BLACKPINK', 'Year in Review'
+    ]
+    assert cipher_signature.keywords == expected
```
RONNCC, line 56: probably would do `expected in cipher_signature.keywords` instead -- sometimes when they re-vamp AI models it adds/remove additional keywords

tfdahlin: For the unit test, we don't really need to worry about this. All of the unit tests use static HTML that's saved in pytube/tests/mocks, so the unit test will always run the same way. 

The unit tests need to use mock HTML, instead of actually querying youtube's servers, because youtube will return 429 (rate limit) status codes when unit tests are run by automated services. 

In the future, if those mocks get updated, it's extremely straightforward to update the test, and needs to be done anyway since the videos have elements that are constantly changing (e.g. views, rating, etc)

PR #835, opened 2020-11-21
Improvement/unavailable video handling

tfdahlin:
There's a special type of video, a members-only video, which is only viewable to users who have subscribed to a content creator (e.g. https://www.youtube.com/watch?v=zcZR2crm8fE). This refactors some of the handling for special cases of unavailable videos by extracting the logic to check if a video is accessible or not into its own function, and introduces a new exception for members-only videos.

This functionality was requested in https://github.com/nficano/pytube/issues/826


tfdahlin:
cc: @RONNCC

pytube/__main__.py
```diff
@@ -102,6 +103,34 @@ def __init__(
             self.prefetch()
             self.descramble()
 
+    def check_availability(self):
```
RONNCC, line 106: nit: check_availability -> check_video_availability ?

pytube/exceptions.py
```diff
@@ -80,5 +80,16 @@ def __init__(self, video_id: str):
         self.video_id = video_id
 
 
+class MembersOnly(PytubeError):
+    """Video is members-only."""
```
RONNCC: nit: clarity.

Video is members-only -> Video is members-only,  a special type of video only viewable to users who have subscribed to a content creator. https://support.google.com/youtube/answer/7544492?hl=en

pytube/extract.py
```diff
@@ -98,6 +98,25 @@ def is_age_restricted(watch_html: str) -> bool:
     return True
 
 
+def playability_status(watch_html: str) -> (str, str):
+    """Return the playability status and reason of a video.
```
RONNCC: Return the playability status and rationale given by Youtube for that status  of a given video. For example a video can be `RecordingUnavailable` with the explanation that `This live stream recording is not available`.

RONNCC (commented):
Random comments to improve code readability.

pytube/__main__.py
tfdahlin: I don't think we really need to use a more verbose function name here. This function should never be called by a consumer of the library, and we don't need to make the code look like java to be readable :p

pytube/extract.py
tfdahlin: I'll update this in a moment

pytube/exceptions.py
tfdahlin: Yeah, I'll make this more descriptive.

pytube/__main__.py
tfdahlin: I'm hesitant to make long, wordy function names because of the fact that we have to take line length into account elsewhere in the code. I don't know if it's the case here, but I have had to split function calls over multiple lines thanks to function and variable names simply being too long, and that significantly detracts from readability imo

pytube/__main__.py
```diff
@@ -102,6 +103,34 @@ def __init__(
...
+        if self.watch_html is None:
+            raise VideoUnavailable(video_id=self.video_id)
+
+        status, messages = extract.playability_status(self.watch_html)
+        for reason in messages:
+            if status == 'UNPLAYABLE':
+                if reason == (
+                    'Join this channel to get access to members-only content '
+                    'like this video, and other exclusive perks.'
+                ):
```
glubsy, line 121: Warning! These reason checks will not work for everybody! These messages are generated based on the location of the IP address of the requesting client. If your IP address has been geo-located by Google/Youtube to be in an english speaking country, it will indeed yield these messages, however, it will change the language if the IP originates from other countries!

pytube/__main__.py
```diff
@@ -102,6 +103,34 @@ def __init__(
...
+                    raise MembersOnly(video_id=self.video_id)
+                elif reason == 'This live stream recording is not available.':
+                    raise RecordingUnavailable(video_id=self.video_id)
+                else:
+                    raise VideoUnavailable(video_id=self.video_id)
+            elif status == 'LOGIN_REQUIRED':
+                if reason == (
+                    'This is a private video. '
+                    'Please sign in to verify that you may see it.'
```
glubsy, line 130: This won't work if the IP is not from an English speaking country!

pytube/__main__.py
```diff
@@ -102,6 +103,34 @@ def __init__(
...
+
+        status, messages = extract.playability_status(self.watch_html)
+        for reason in messages:
+            if status == 'UNPLAYABLE':
+                if reason == (
+                    'Join this channel to get access to members-only content '
+                    'like this video, and other exclusive perks.'
+                ):
+                    raise MembersOnly(video_id=self.video_id)
+                elif reason == 'This live stream recording is not available.':
```
glubsy, line 123: This won't work if the IP is not from an English speaking country!

pytube/__main__.py
tfdahlin: https://github.com/pytube/pytube/blob/4faaa5fe8d88abd7b6fbaaab5dc287f0a3f5c502/pytube/request.py#L19
We have this accounted for in the way that we fetch videos -- we explicitly tell YouTube to give us english-language results for this.

pytube/__main__.py
tfdahlin: See above

pytube/__main__.py
tfdahlin: See above

pytube/__main__.py
glubsy: Awesome! I did not notice that. Nevermind then. Thanks! :)

PR #838, opened 2020-11-21
Improvement/parser

### LINKED ISSUE #836: JavaScript object parser
Currently, there are a number of extract functions that rely on regular expressions to find the end of the object. As an example, the HTML of a page has code that looks like this:
```javascript
ytplayer.config = {"args":{"js": "path/to/file/base.js","initial_response":"lots of stuff in here"}};
```

At present, pytube uses regexes to identify these patterns, but this doesn't always work. For example, to find the above assignment, the regex 
```python
pattern = r"ytplayer\.config\s*=\s*({.+?});"
```
was used for a long time. However, this can break pretty easily. We have seen videos that have unusual string in them, used for templating (e.g. "url/{gdpr};") which causes the regular expression to find a result too early. This might be possible to work around if python's regex library supported overlapping matches when using findall(), but unfortunately it does not.

To more effectively solve this problem, we will need to build a better function for parsing the javascript code. This code *probably* won't be too complex, because it will only ever be parsing objects, so it will mostly be a matter of tracking opening/closing brackets, braces, and quotation marks. Writing this will almost certainly allow pytube to more robustly parse objects when there are new changes to YouTube, but only if it's done well.

If this is implemented correctly, we may need to switch from using `json.loads()` to using `ast.literal_eval()`, and may need to make inline changes to the data being processed. If ast is used instead of json.loads, an example of inline-modification that will need to be handled is converting boolean strings from all lowercase to being capitalized, and in the future we may need to handle special cases, such as converting `NaN` to `float('NaN')`. These special cases are unlikely to exist in the code being parsed, but may be needed at a later date, so extensibility in the design of the parser should be taken into consideration.

tfdahlin:
This PR implements new functionality that improves pytube's parsing capabilities for javascript objects.

For more details, see #836 and the discussion in [this PR](https://github.com/nficano/pytube/pull/808).

### DIFF
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -18,10 +18,12 @@ from urllib.parse import unquote
 from urllib.parse import urlencode
 
 from pytube.cipher import Cipher
+from pytube.exceptions import HTMLParseError
 from pytube.exceptions import LiveStreamError
 from pytube.exceptions import RegexMatchError
 from pytube.helpers import regex_search
 from pytube.metadata import YouTubeMetadata
+from pytube.parser import parse_for_object
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


tfdahlin:
@a-czyrny would love your input on this. I've tested it pretty extensively, and I'm not running into any issues with it locally, but would appreciate another pair of eyes and some additional testing since YouTube behaves strangely sometimes.

@RONNCC i'd appreciate your input on these changes as well if you get the chance

pytube/extract.py
```diff
@@ -269,31 +271,29 @@ def get_ytplayer_config(html: str) -> Any:
     """
     logger.debug("finding initial function name")
     config_patterns = [
-        r"ytplayer\.config\s*=\s*({.+?});ytplayer",
-        r"ytInitialPlayerResponse\s*=\s*({.+?(?<!gdpr)});"
+        r"ytplayer\.config\s*=\s*",
```
RONNCC, line 274: usually ? after *  to choose the shortest match or it can get really really long if an extra character gets added

pytube/parser.py
```diff
@@ -0,0 +1,83 @@
+import ast
+import json
+import re
+from pytube.exceptions import HTMLParseError
+
+
+def parse_for_object(html, preceding_regex):
+    """Parses input html to find the end of a JavaScript object.
```
RONNCC, line 8: can we add where we have a need to parse the object? 
if i remember correctly it's because youtube literally sends javascript as html so you need to parse it out?

pytube/parser.py
```diff
@@ -0,0 +1,83 @@
...
+    :returns:
+        A dict created from parsing the object.
+    """
+    html = html[start_point:]
+    if html[0] != '{':
+        raise HTMLParseError('Invalid start point.')
+
+    # First letter MUST be a open brace, so we put that in the stack,
+    # and skip the first character.
+    stack = ['{']
```
RONNCC, line 44: hmm why not use the regex instead of the while loop here? bit confused as to the purpose / additional value

RONNCC:
thoughts on matching `r"ytplayer\.config\s*={.+);\n"` or something? (i.e. only match the ending semicolon and the ytplayer.config= beginning - not sure if that would work well as haven't tried :)

pytube/extract.py
tfdahlin: I would agree if we were matching with `\.*`, but we're just matching whitespace, and in the case of the second `\s*`, we have to match *all* whitespace preceding the `{` character. In the actual string we're searching, to the best of my knowledge, this always appears in the text as `ytplayer.config = {<stuff here>}`. If we were to use, say `ytplayer\.config\s*?=\s*?`, the space preceding the actual object would not be captured, and `parse_for_object` would fail.

pytube/parser.py
tfdahlin: See https://github.com/nficano/pytube/issues/836

tfdahlin:
I have an explanation of the issue with simply using a regex in https://github.com/nficano/pytube/issues/836

pytube/parser.py
tfdahlin: Yeah, we're parsing the javascript portion of the html page for a video to extract information. In some cases, the javascript sets variable values as long, descriptive objects, and we *were* using simple regex patterns to try and find these, but that becomes... problematic pretty quickly. I gave some examples where a normal regex fails to accurately capture the object because it terminates too early, and python regexes don't support overlapping matches (meaning we can't use `findall()` and iterate through results until we find a correct match).

tfdahlin:
There's also some additional context in [this comment](https://github.com/nficano/pytube/pull/808#issuecomment-728625599) in a previous PR.

PR #842, opened 2020-11-23
Improvement/reduce network calls

### LINKED ISSUE #841: Question: example for multiple calls over longer time
Hi, I'd like to reuse this project in a Mediaplayer. 
So over the time it'll have to decrypt a lot of videoid to stream urls.

Is there any example that divides one-time needed calls, like loading the decrypt algorithm from the per-video part, where the decryption takes place? 

Thanks a lot!!

tfdahlin:
Introduces new cached variables to the library, `pytube.__js__` and `pytube.__js_url__`, in order to reduce the number of network requests made by an application. The media player javascript typically won't change during the course of an application run, so there's no reason to download the javascript for *every* video. This should speed up the code by a second or two here or there on longer runs that download multiple videos.

### DIFF
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -14,6 +14,7 @@ from typing import List
 from typing import Optional
 from urllib.parse import parse_qsl
 
+import pytube
 from pytube import Caption
 from pytube import CaptionQuery
 from pytube import extract
@@ -137,12 +138,6 @@ class YouTube:
                 apply_descrambler(self.vid_info, fmt)
             apply_descrambler(self.player_config_args, fmt)
 
-            if not self.js:
-                if not self.embed_html:
-                    self.embed_html = request.get(url=self.embed_url)
-                self.js_url = extract.js_url(self.embed_html)
-                self.js = request.get(self.js_url)
-
             apply_signature(self.player_config_args, fmt, self.js)
 
             # build instances of :class:`Stream <Stream>`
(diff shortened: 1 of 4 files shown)


tfdahlin:
cc: @RONNCC

tfdahlin:
We should *probably* do a minor version bump when this, #835 and #838 get merged

tfdahlin:
cc: @RONNCC for this, #835, and #838

p.s. hope you had a good vacation!

RONNCC:
@tfdahlin  hope your Thanksgiving 🦃  was good! I know I definitely gained a few pounds haha 😄

tfdahlin:
It wasn't bad! Can't wait to be able to spend it with friends and family next year though

pytube/__main__.py
```diff
@@ -185,18 +180,26 @@ def prefetch(self) -> None:
...
         self.initial_data_raw = extract.initial_data(self.watch_html)
         self.initial_data = json.loads(self.initial_data_raw)
 
         self.vid_info_raw = request.get(self.vid_info_url)
-        if not self.age_restricted:
-            self.js_url = extract.js_url(self.watch_html)
+
+        # If the js_url doesn't match the cached url, fetch the new js and update
+        #  the cache; otherwise, load the cache.
+        if pytube.__js_url__ != self.js_url:
```
RONNCC, line 197: think we could use https://github.com/nficano/pytube/blob/master/pytube/helpers.py#L113 instead?

tfdahlin: That can't really be used in this case.

The goal of this fix is to prevent repeated network requests to download the `base.js` file from YouTube across a much longer-running application. That helper would only work on a per-object basis, so every new YouTube object will still make the network request.

My implementation caches at the module level -- so every new object will have access to the cached information.

tfdahlin: so, for example, lets say you have 5 different videos that you want to download:

```python
my_videos = [
   # video links in here
]

for video in my_videos:
    yt = YouTube(video)
    yt.streams.first().download()
```

The line that instantiates the YouTube object will usually *always* fetch the base.js. 
My fix implements a module-level variable that's checked before attempting to download the script. Unless the URL changes for the js file, it will only download the js once.
Using that helper function only caches the function call on a specific object, so it will still download the base js 5 times.

RONNCC: Is it safe to say - this is downloading media player from yt with assumption static file wont change in a short duration?

What would happen for long running scripts (?)  e.g. ```for (;;) { download video from $1 }```

tfdahlin: To some extent, yes. It assumes that the static file won't change in a short duration while also retaining the same url.

It's worth noting that the js url take the form of `/s/player/<random?string>/player_ias.vflset/en_US/base.js`. Currently, whenever I load a page, the random string is 408be03a, which I believe is a sort of hash of the file. 

Any time that the watch_html or embed_html gets loaded, it refers back to the js url to see if the url has changed, and will update the cache if it has.

tfdahlin: Given the fact that any fundamental changes to this js file will probably break pytube altogether until [extract](https://github.com/nficano/pytube/blob/master/pytube/extract.py#L327) gets updated, I think this is a pretty safe thing to cache.

tfdahlin: cc: @RONNCC

RONNCC: ah sg

tfdahlin:
@RONNCC can we get a minor version bump + push to pypi?

PR #871, opened 2020-12-10
Github workflow for publish to PyPI

tfdahlin:
This creates a GitHub workflow to automate pushes to PyPI.

@RONNCC this will require creation of a PyPI API token, which can be scoped exclusively to the pytube repository. See the `Saving Credentials` section on [this page](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) for an explanation. This workflow does not publish to the test PyPI, so we do not need those credentials stored in the repository secrets.

This will *only* run on tagged commits (per line 45) that are pushed to `master` (per line 6).


RONNCC:
Ok let me look into this over the weekend

tfdahlin:
Did you get the chance to look at this @RONNCC ?

RONNCC:
Probably going to look at it Friday @tfdahlin  -- flooded by last minute work

tfdahlin:
Aight, sounds good

tfdahlin:
I added a couple of conditional checks to both the pypi publishing workflow, as well as the github release workflow, so that dependabot updates won't trigger them.

RONNCC:
@tfdahlin actually just realised  - I don't have permissions to do this on the github repo 🤔 
(i.e. I cant see the settings page to add tokens there)

tfdahlin:
@RONNCC oh. I guess we'll need to get nick involved then

tfdahlin:
@RONNCC I emailed nick about getting that secret set up. In the meantime, can you push 10.1.0 to pypi please?

polarAli:
Hi, please publish v10.1.0 the #872 bug fix is necessary.

RONNCC:
@polarAli  -- done; you can always download the latest from github by using the git link in your pip install

polarAli:
@RONNCC I very much appreciate your help.

tfdahlin:
I was way overcomplicating this with my initial implementation. What I've done is removed the GitHub release workflow (which was broken anyway, I think because of a token rotation or something) in favor of this. Now we can manually create new releases on GitHub, which on creation will trigger an update on pypi.

tfdahlin:
@RONNCC I've added a pypi token as a repository secret, just want a second pair of eyes to take a peek at this before I merge it. This should simplify the pypi deploy process significantly

RONNCC:
sgtm @tfdahlin

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,32 @@
...
+          python-version: 3.8
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install build setuptools wheel twine
+
+      - name: Build and publish
+        env:
+          TWINE_USERNAME: __token__
```
RONNCC, line 28: does it bump version somewhere ...?

tfdahlin: @RONNCC this workflow will only trigger when a GitHub release is made, which can be done from [this page](https://github.com/pytube/pytube/releases) by admins. Our workflow for this would be to merge a PR, use `bump2version` to handle version bumping, then come to the repo and create the release. 

I'm hesitant to use an automatic release creation because I saw it getting triggered by dependabot a lot in... weird ways. Since pytube doesn't have any direct dependencies (the pipfile only has dev-packages, and its all linting stuff), any time a dependency it didn't rely on had a minor/major version update, it would create minor/major release for pytube which doesn't make sense since nothing actually changed for our users.

tfdahlin: I don't think it makes sense to have automated version bumping unless we get it right the very first time, because I think it would be easy to accidentally make major version releases when we really should only do minor/patch releases

RONNCC: ah i see ok

Issue #873, opened 2020-12-13
[BUG] on_progress_callback & on_complete_callback arguments


maxouille:
Hi, 

I'm using the 10.1.0 version of pytube. 

When I try to add progress callback, it takes 3 args : `chunk: bytes, file_handler: BinaryIO, bytes_remaining: int`. However, the three args i'm using have types `Stream, bytes, int` and they are passed from the YoutTube object. 

Moreover, the on_complete callback is not working, the app crashes with a random negative exit code.

Here is my code : 
```
# This works good but not coherant with doc
def progress_function(stream, chunk, bytes_remaining):
    total_size = stream.filesize
    bytes_downloaded = total_size - bytes_remaining
    percentage_of_completion = bytes_downloaded / total_size * 100
    window.ui.progressBar.setValue(int(percentage_of_completion))

# This makes the app crash. The print is not even written on the console.
def completed_function():
    print("download conmpleted")
    window.ui.progressBar.setVisible(False)
    window.ui.info_text.setText("Téléchargement terminé !")
 
youtube = pytube.YouTube(self.url_link)
# Set callbacks for on going download and completed download
youtube.register_on_progress_callback(progress_function) # warning Expected type 'OnProgress', got (stream: {filesize}, chunk: Any, bytes_remaining: Any) -> None instead
youtube.register_on_complete_callback(completed_function) # warning Expected type 'OnComplete', got () -> None instead
video = youtube.streams.get_highest_resolution()
dir_path = QFileDialog().getExistingDirectory(self, 'Save MP4 ...', expanduser('~'), QFileDialog.ShowDirsOnly)
print("Download will start...")
video.download(dir_path)
```

I'm using python 3.9 and installed pytube from the UI of PyCharm.

**EDIT : 
It looks like I messed up with the on_progress method on the Stream Object and the register_on_progress_callback method on the Youtube Object ! However, would you please explain to me how to fix the warnings described above ?** 

Thanks a lot !

tfdahlin:
Hi @maxouille, this is actually an issue where we're just lacking good documentation about how to use the library. The correct way to register on-progress and on-complete callbacks is actually in the YouTube object initialization.

See [this comment](https://github.com/nficano/pytube/issues/862#issuecomment-740014886) from another issue for an explanation of how to use on_progress and on_complete. I believe this will fix the issues you're running into!

maxouille:
Thank you, I managed to make it work, however, I still have warnings : 

```
[...]

PR #895, opened 2020-12-28
Type annotation fixes

tfdahlin:
Attempts to fix an issue with #873 where type annotations were incorrect and raising warnings in pycharm.


tfdahlin:
cc: @RONNCC

Issue #874, opened 2020-12-13
Add documentation for registering on_progess and on_complete callbacks


tfdahlin:
The current documentation for registering on_progress and on_complete callbacks is incorrect, I believe due to changes from the pytube3->pytube merge. We've had a couple of people ask about how to use this, because the README says one thing, but in reality, this doesn't work correctly. This also isn't documented in readthedocs, so it should be added there as well.

See [this comment](https://github.com/nficano/pytube/issues/862#issuecomment-740014886) for a description of how to implement on_progress and on_complete callbacks.

tfdahlin:
cc: @sneha-001 if you want something specific that needs documentation, here's one you could take a stab at to test the waters

sneha-001:
@tfdahlin Sounds good! I'll take over for writing the documentation for those!

maitreyee19:
@tfdahlin I have checked in the documentation change . I think this can be closed.

tfdahlin:
@maitreyee19 you've updated the readme, but we still need the files for readthedocs to get updated at some point, so for now this issue will remain open. I'll probably write that documentation myself later this week

maitreyee19:
@tfdahlin  if you have not updated the readthedocs, I can take it up. I also observed that some of the old readme changes are not merged to readthedocs.

tfdahlin:
@maitreyee19 I actually have a branch going for a pretty big documentation update, it's just taking me a little while to finish it. Hopefully going to have it done by the weekend

PR #911, opened 2021-01-10
Readme Update for #874

maitreyee19:
documentation for registering on_progess callbacks.

Tried to keep the change in documentation to minimal . Only fixing the problem with the documentation . Didn't add documentation to create the call back using the constructor.


maitreyee19:
Also added document to readme about for downloading file to specific file and folder . I was looking for this documentation for few days . Then I went through the code to find the usage (and used it myself). Sure it will help some new users to pytube.

tfdahlin:
Looks good to me

PR #924, opened 2021-01-31
Improvement/docs

### LINKED ISSUE #874: Add documentation for registering on_progess and on_complete callbacks

tfdahlin:
Lot of doc changes here. Notably, removed a lot of the fluff from the README in favor of putting it into actual doc files for readthedocs, with a link pointing to that documentation. Also moved bumpversion and twine from the package requirements to the dev-package requirements so pytube remains dependency-free.

cc: @RONNCC


tfdahlin:
@RONNCC @nficano either of you have a moment to provide some input on this? Don't merge this yet, just want your thoughts on extracting the detailed usage information from the README into readthedocs, and linking to that instead, so we have less of a wall of text there.

tfdahlin:
https://github.com/tfdahlin/pytube/tree/improvement/docs you can check out what the README will look like after these changes over here

RONNCC:
lgtm - would love if we could call out the documentation link more somehow 🤔  but not sure how hmm

tfdahlin:
The link already exists in the about section on the side of the GitHub page, but we *could* move it up to the very beginning section of the README so it's not linked at the bottom if you think that would look better

tfdahlin:
Updated the README, hopefully it makes the documentation link a little more visible.

tfdahlin:
Gonna go ahead and merge this. We can update it later as needed

RONNCC:
sgtm

PR #925, opened 2021-02-02
Fix : getting non-progressive streams by resolution

itsachrafmansari:
I think users should be able to get streams with a certain resolution even if they aren't progressive mp4.
And by manually using the filter() function, they can specify which ones they want to get.

Example :
```Python
from pytube import YouTube

vid = YouTube(r"https://www.youtube.com/watch?v=qmtmse1xh_Q")

print(vid.streams.get_by_resolution("480p"))
```

Before, the output was :
> None

Now it's :
> <Stream: itag="135" mime_type="video/mp4" res="480p" fps="30fps" vcodec="avc1.4d401f" progressive="False" type="video">

***
We can manually specify what streams we want to get with the filter() function :
```Python
print(vid.streams.filter(progressive=True, subtype="mp4").get_by_resolution("480p"))
```
So we will get :
> None

### DIFF
--- a/pytube/query.py
+++ b/pytube/query.py
@@ -253,8 +253,6 @@ class StreamQuery(Sequence):
     def get_by_resolution(self, resolution: str) -> Optional[Stream]:
         """Get the corresponding :class:`Stream <Stream>` for a given resolution.
 
-        Stream must be a progressive mp4.
-
         :param str resolution:
             Video resolution i.e. "720p", "480p", "360p", "240p", "144p"
         :rtype: :class:`Stream <Stream>` or None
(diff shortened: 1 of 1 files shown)


tfdahlin:
Hi @itsachrafmansari you actually *can* filter by resolution really easily:
 ```python
>>>  vid.streams.filter(resolution="480p")
[<Stream: itag="135" mime_type="video/mp4" res="480p" fps="30fps" vcodec="avc1.4d401f" progressive="False" type="video">,
<Stream: itag="244" mime_type="video/webm" res="480p" fps="30fps" vcodec="vp9" progressive="False" type="video">,
<Stream: itag="397" mime_type="video/mp4" res="480p" fps="30fps" vcodec="av01.0.04M.08" progressive="False" type="video">]
 ```
 
 The `get_by_resolution()` function is just a convenient shortcut. The main purpose for this shortcut is for the CLI, so it really shouldn't be used in code. I'll look into some changes to make this more clear though.

itsachrafmansari:
Oh! How did I totally forget about that :man_facepalming:

Thanks for your reply, I have just realised the mistake I did.

PR #932, opened 2021-02-13
Initial implementation of Channel object

### LINKED ISSUE #909: Is is possible to iterate over a channels videos ?
I can't seem to find the Channel object and the Channel.videos() method

### LINKED ISSUE #964: Channel URL
`pytube.YouTube` extracts many useful details from a video list, it would be good to extract channel URL as well.
for example, python API for  `youtube_dl` also supports this feature

tfdahlin:
Pulled a lot of the implementation from the playlist class, this should allow users to iterate over a channel's videos.

Still need to write unit tests and flesh it out a bit, but it at least seems to sorta work right now


tfdahlin:
Something I noticed while writing this is that the `.video_urls` property of the playlist object is extremely inefficient if a user only wants a small subset off the URLs. I'm going to try and improve the way this works before I even look at merging this, so users can make less network calls to access a list of urls.

tfdahlin:
I still need to write unit tests, which I'll do a little later. @RONNCC, I would appreciate your eyes on this.

Summary of commits (since there have been a lot):
* Implements a list-like object, `DeferredGeneratorList`
  * This has some wrapper functionality to convert generators into lazy-loading lists. The benefit of this is apparent performance in specific use cases. For example, when downloading a very large playlist, pytube currently populates the ENTIRE youtube list, which incurs a number of web requests as it generates the initial list before it can start downloading. This offsets these requests to only happen as needed. User experience should see a positive impact (instead of waiting for a long series of requests, the user will more quickly begin processing, and only request as needed). This also helps with testing because I can e.g. check if a single continuation works by using list comprehensions.
* Makes minor changes to the `Playlist` class so that it can be subclassed for other purposes. Channels and playlists behave very similarly for API interaction stuff, so I didn't want to rewrite much of the code. In the future, we may look at abstracting some of the functionality, but it seems a little unnecessary for now.
* Adds a `Channel` class for more cleanly interacting with user channels. Right now the work-around people use is to select the "Watch all videos" from a user's channel to get a playlist of all the videos, but this will be more intuitive, and could be used to e.g. navigate from a video to a channel more easily in the future.

tfdahlin:
Unit tests written, ready to merge if it looks good to you cc: @RONNCC

PR #938, opened 2021-02-23
Support for quoted attributes in transform object

guillaume-plc:
A few hours ago all of the sudden I couldn't download anything and systematically got hit with an error when trying to load a url:
```
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    yt = YouTube(url)
  File "C:\Users\user\envs\ytgui\lib\site-packages\pytube\__main__.py", line 105, in __init__
    self.descramble()
  File "C:\Users\user\envs\ytgui\lib\site-packages\pytube\__main__.py", line 177, in descramble
    apply_signature(self.player_config_args, fmt, self.js)
  File "C:\Users\user\envs\ytgui\lib\site-packages\pytube\extract.py", line 370, in apply_signature
    cipher = Cipher(js=js)
  File "C:\Users\user\envs\ytgui\lib\site-packages\pytube\cipher.py", line 37, in __init__
    var, _ = self.transform_plan[0].split(".")
ValueError: not enough values to unpack (expected 2, got 1)
```

So I looked a bit inside cipher.py and it seems that YouTube slighlty modified its cipher syntax: some transform functions are now stored inside quoted attributes of the obfuscated variable, accessed via brackets:

```javascript
var Mx={
  FH:function(a){a.reverse()},
  "do":function(a,b){var c=a[0];a[0]=a[b%a.length];a[b%a.length]=c},
  xK:function(a,b){a.splice(0,b)}
};
```

A typical transform_plan now looks something like this, which breaks the routine to retrieve the obfuscated variable name, and the parse_function method.
```python
['Mx["do"](a,17)', 'Mx.FH(a,61)', 'Mx.xK(a,3)', 'Mx["do"](a,12)', 'Mx.xK(a,1)', 'Mx.FH(a,37)', 'Mx["do"](a,47)', 'Mx.FH(a,6)']
```

Fortunately the transform map is still built correctly, so I just replaced the routine to find the variable name with a short regex and modified the parse_function method to allow for multiple patterns, and it seems to work now !

PR #939, opened 2021-02-23
Fixed error because transform_plan does not use dot but bracket

minsukkahng:
The pytube.YouTube function creates a ValueError because transform_plan has a new pattern. 
Error message: `ValueError: not enough values to unpack (expected 2, got 1)`

This is because transform_plan does not always use ".", but it uses "[..]". An example of self.transform_plan includes:
`['Mx["do"](a,17)', 'Mx.FH(a,61)', ...]`

This PR checks if self.transform_plan[0] contains either "." or "[".

Thank you.

Cheers,
Minsuk

### DIFF
--- a/pytube/cipher.py
+++ b/pytube/cipher.py
@@ -33,8 +33,17 @@ logger = logging.getLogger(__name__)
 class Cipher:
     def __init__(self, js: str):
         self.transform_plan: List[str] = get_transform_plan(js)
-        var, _ = self.transform_plan[0].split(".")
+
+        if "." not in self.transform_plan[0] and "[" not in self.transform_plan[0]:
+            raise RegexMatchError(
+                caller="__init__", pattern="multiple"
+            )
+        elif "." in self.transform_plan[0]:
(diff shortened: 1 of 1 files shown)


minsukkahng:
Another PR seems to deal with the same issue (https://github.com/pytube/pytube/pull/938). Please feel free to reject mine.

guillaume-plc:
Hey I see we've faced the same issue, did you handle the parse_function method ? For me only fixing the initialization was not enough as the brackets also break the js_func_regex pattern !

minsukkahng:
> Hey I see we've faced the same issue, did you handle the parse_function method ? For me only fixing the initialization was not enough as the brackets also break the js_func_regex pattern !

Hi @guillaume-plc, 

Yes, we've got the same issue. It was fine for me to use my string match approach, but I realized your approach would work better.

tfdahlin:
Hey, gonna close this one in favor of @guillaume-plc's PR, just going to do a once-over on that first.

PR #952, opened 2021-03-07
Consolidate import sections

tfdahlin:
This makes a few changes to how imports are written in pytube. 

Take [\_\_main\_\_.py](https://github.com/pytube/pytube/blob/master/pytube/__main__.py#L10-L36) for example, which has 26 lines of imports. Some of the flake8 rules require that every import have its own line, so as the project expands, these lines of imports are likely to become longer and longer. By excluding the rules that prohibit * imports, and having multiple imports on the same line (e.g. `import os, json`), we can have much shorter import sections of these files.

cc: @RONNCC @nficano thoughts?


RONNCC:
hmm --  think pep8 prefers lots of lines?
https://stackoverflow.com/questions/24271961/is-there-anything-bad-about-having-multiple-imports-on-one-line

tfdahlin:
PEP8 might prefer lots of lines, true, but at a certain point it becomes ridiculous. If I have to scroll a full screen length to get to meaningful code, then something is wrong.

If we didn't already have a lot of sanitation checks in our code (e.g. flake8 checks for unused imports), I might be more concerned with following PEP8 exactly, but I think this is an easy win for us to make the code more accessible

tfdahlin:
https://youtu.be/FyCYva9DhsI?t=1676 found the video that influences my opinions on this, timestamp for the relevant part

RONNCC:
hmm personally I prefer the imports because it's clearer when reading code online rather than the import * equivalent 😄 

otherwise i need to wait for an index to build before i can do 'goto'

tfdahlin:
alright, well, i'll revert to pep8 standards, but i'm still going to squash a bunch of stuff since we dont need a line for each individual function/object imported from a package

Issue #922, opened 2021-01-29
Get all videos titles from youtube using pytube


christianfelipesf:
**Hello I made this small change in the script that can harvest all the titles of a video from the pytube directly from the url call files, as I still haven't learned to download all the pages after 100, I had to make changes to the main file, it helped me to take all titles and have several uses mainly for graphical interfaces and also have control of when a video in your music playlist was private or deleted**

https://github.com/edergames29/pytubetitleplayslitfast

ghw2020:
A simple way to get a playlist containing all videos of a channel is described in:

https://github.com/pytube/pytube/issues/619#issuecomment-774460819

PR #953, opened 2021-03-07
Defer web requests until absolutely necessary

### LINKED ISSUE #951: Defer web requests by default until necessary for accessing information
**Is your feature request related to a problem? Please describe.**
Not necessarily related to a problem, but could lead to performance improvements

**Describe the solution you'd like**
There are a lot of moving pieces in YouTube objects, especially with the way their properties are accessed. My goal is to improve performance by deferring web requests wherever possible. For example, in order to access YouTube object titles, a web request to the video's url must be made. However, in some cases this information might already be available without initiating any web requests. For example, when browsing a playlist, there is often a lot of metadata about videos already available without loading the entire video, such as the titles, creators, and length. If we defer requests for that data unless it's necessary, then we can add the metadata that has been fetched elsewhere to reduce the number of network requests being made.

**Describe alternatives you've considered**
N/A

**Additional context**
#922 -- fetching youtube video titles from playlists faster

tfdahlin:
This defers a lot of the network requests until they're absolutely necessary. For example, there's no real reason to fetch the embed_html or js file if the user is only trying to get the title, which should be available from watch_html. Additionally, this will pave the way for pre-filling some aspects of a YouTube object for operations on non-video elements, like playlists and channels. Currently, if a user tries to iterate over `Playlist.videos` and comes across a video that is not accessible, an uncatchable exception will be thrown, because it is being thrown in a generator. By deferring this kind of thing until later, the user will be able to catch these issues inside of a loop, rather than needing to create a loop calling next(). Additionally, this will allows us to set certain attributes of a YouTube object without needing to make any web requests at all sometimes. For example, playlists should be able to access the video title without needing to make a request to every single video page to do so.

When merged, this closes #951

### DIFF
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -76,12 +76,12 @@ class YouTube:
 
         self._watch_html: Optional[str] = None  # the html of /watch?v=<video_id>
         self._embed_html: Optional[str] = None
-        self.player_config_args: Dict = {}  # inline js in the html containing
-        self.player_response: Dict = {}
+        self._player_config_args: Optional[Dict] = None  # inline js in the html containing
+        self._player_response: Optional[Dict] = None
         # streams
         self._age_restricted: Optional[bool] = None
 
-        self.fmt_streams: List[Stream] = []
+        self._fmt_streams: Optional[List[Stream]] = None
 
         self._initial_data = None
         self._metadata: Optional[YouTubeMetadata] = None
@@ -99,9 +99,6 @@ class YouTube:
 
         if proxies:
             install_proxy(proxies)
-
-        if not defer_prefetch_init:
-            self.descramble()
 
     @property
     def watch_html(self):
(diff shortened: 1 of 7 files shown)


tfdahlin:
@nficano @RONNCC thoughts on this change? I feel like if we defer any web requests until the last minute, and cache the results of those requests, we can facilitate alternative program flows that some users may have

pytube/__main__.py
```diff
@@ -177,80 +239,60 @@ def descramble(self) -> None:
...
-        Take the unscrambled stream data and uses it to initialize
-        instances of :class:`Stream <Stream>` for each media stream.
+        status, messages = extract.playability_status(self.watch_html)
 
-        :param str fmt:
-            Key in stream manifest (``ytplayer_config``) containing progressive
-            download or adaptive streams (e.g.: ``url_encoded_fmt_stream_map``
-            or ``adaptive_fmts``).
+        for reason in messages:
+            if status == 'UNPLAYABLE':
```
RONNCC, line 265: can we enum values /w a typeclass or similar :) 
general  string constants scare me

RONNCC: the status and reason since they both seem like they can be mostly fixed? 
no clue

RONNCC:
@tfdahlin  could you elaborate on some sort of web flows that may fit this pattern?

tfdahlin:
I think the easiest example is https://github.com/pytube/pytube/issues/922 (since it's a real user's use case) -- they're just trying to extract video titles from a playlist.

Currently, the playlist object only tracks the URLs for its videos, even though there is additional information available to us when we parse the page, such as their titles. We *could* create a list of titles as part of the playlist object, but what if there's other information we want to extract, like the account that posted it? We *could* create a new class for video information, but what's the point when we have YouTube objects that already exist, and those objects have properties/attributes for all the information we might extract?

I also wrote another example of why it might be useful [here](https://github.com/pytube/pytube/issues/951#issuecomment-792227284).

By deferring web requests until we actually need to fetch information, this could save significant runtime for certain workflows, especially if we make certain attributes of a `YouTube` object settable, like the title/author (which I've done in this PR).

pytube/__main__.py
tfdahlin: The status and reason strings are extracted directly from youtube, and are not constants that I randomly decided on

tfdahlin:
cc: @RONNCC

RONNCC:
Hmm 🤔  generally prefer webcalls done once rather than lazily in background if I'm understanding correctly but i dont have strong objection -- feel free to merge

PR #956, opened 2021-03-12
Timeout Patch [Temporary]

NetherStar64:
YouTube must have changed something, because now creating a new YouTube Object with pytube takes very long. 
My Solution to this is to add a timeout parameter to when requesting, but this is not a very good solution and is only temporary.

Also sorry for the many unnessesary commits

### DIFF
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -205,7 +205,7 @@ class YouTube:
 
         if self.age_restricted:
             if not self.embed_html:
-                self.embed_html = request.get(url=self.embed_url, timeout=2)
+                self.embed_html = request.get(url=self.embed_url)
             self.vid_info_url = extract.video_info_url_age_restricted(
                 self.video_id, self.watch_url
             )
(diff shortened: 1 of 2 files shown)


tfdahlin:
Is this an issue you're having with all videos, or only specific ones?

NetherStar64:
> Is this an issue you're having with all videos, or only specific ones?

No it isn't working on all Videos. It takes like 5 Minutes to finish ```vid = YouTube("Link") ``` on multiple devices, but with the timeout it works perfectly fine

tfdahlin:
OH. I see what the problem you're having is. I thought the problem you were running into was downloading a video, not creating a YouTube object. I actually have a much more comprehensive solution for your problem in #953. Adding a timeout the way that you have will most likely break downloads, even if it makes the objects get created faster.

Basically what's going on behind the scenes with YouTube objects now is they're greedily making 2-3 web requests, which slows down object creation. The PR that I linked defers those web requests until they're necessary, so that you can create objects quicker, and it will only ever make the web requests when it needs to access information.

I'll be merging that soon hopefully, I'm just waiting for at least one of the other maintainers to take a look at it first to get their input.

tfdahlin:
Since I will be merging my changes in #953 soon, I'm going to close this PR

PR #960, opened 2021-03-18
Link corresponding website to all badges

jcs090218:
This patch does the following.

- Organize markdown headers for readability
- Link all badges to it's reasonable website


tfdahlin:
Looks good to me

PR #965, opened 2021-03-28
really remove typing_extensions requirement

dotlambda:
setup.py was forgotten by https://github.com/pytube/pytube/pull/801


tfdahlin:
Good catch, ty!

PR #966, opened 2021-04-03
Enhancement/timeout

### LINKED ISSUE #947: Network problems cause Downloads to freeze
When downloading any video (or audio for that matter), when my computer experiences network problems such as sudden disconnects or reconnects to other networks, the download just freezes. The file does no longer update, its size does not grow and the script seems to have stopped making any requests to the network (looked it up with a network monitor). Over a span of 1h after my pc got disconnected from the network (reconnected 5 minutes later), the video file did not grow/update, making it seem like the download is just frozen. 
I use multiprocessing process to handle downloads. When killing the script, the following error message appears

  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/me/ytscript.py", line 1986, in download
    vid.download(output_path=sys.path[0], filename=in_vid)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/pytube/streams.py", line 260, in download
    for chunk in request.stream(self.url):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/pytube/request.py", line 112, in stream
    chunk = response.read(chunk_size)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/http/client.py", line 447, in read
    n = self.readinto(b)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/http/client.py", line 491, in readinto
    n = self.fp.readinto(b)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/socket.py", line 589, in readinto
    return self._sock.recv_into(b)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 1049, in recv_into
    return self.read(nbytes, buffer)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 908, in read
    return self._sslobj.read(len, buffer)
KeyboardInterrupt

it seems like a timeout somewhere would be a good idea

tfdahlin:
Adds optional request timeout and max retry arguments to `streams.download()`. If a network connection gets closed for unforseen reasons, pytube may get stuck attempting to download something it is unable to download. This allows users to set a timeout and retry policy for such downloads.

Closes #947

### DIFF
--- a/pytube/exceptions.py
+++ b/pytube/exceptions.py
@@ -13,6 +13,10 @@ class PytubeError(Exception):
     """
 
 
+class MaxRetriesExceeded(PytubeError):
+    """Maximum number of retries exceeded."""
+
+
 class HTMLParseError(PytubeError):
     """HTML could not be parsed"""
 
--- a/pytube/request.py
+++ b/pytube/request.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 """Implements a simple wrapper around urlopen."""
+import json
 import logging
 from functools import lru_cache
 import re
-import json
+import socket
 from urllib import parse
 from urllib.request import Request
 from urllib.request import urlopen
@@ -16,7 +17,13 @@ default_chunk_size = 4096  # 4kb
 default_range_size = 9437184  # 9MB
 
 
-def _execute_request(url, method=None, headers=None, data=None):
+def _execute_request(
+    url,
+    method=None,
+    headers=None,
+    data=None,
+    timeout=socket._GLOBAL_DEFAULT_TIMEOUT
+):
     base_headers = {"User-Agent": "Mozilla/5.0", "accept-language": "en-US,en"}
     if headers:
         base_headers.update(headers)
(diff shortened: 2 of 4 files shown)


tfdahlin:
I still need to write unit tests, I'll convert this to a real PR once that's done.

tfdahlin:
cc: @RONNCC

RONNCC:
makes sense -- 

might be  easier to just use the retry/max timeout framework already in  https://docs.python-requests.org/en/master/ ?

tfdahlin:
@RONNCC we don't actually use the python requests library (b/c of the dependency-free goal), we use python's native urllib, which doesn't have that option

PR #974, opened 2021-04-09
Remove "harold.martin@gmail.com" from authors

nficano:
@RONNCC @tfdahlin mind if I add you both to the authors? 

Tangentially related, I purchased the domain pytube-projects.com and I'd like to change the project email to something like hello@pytube-projects.com, which would forward to the three of us, would that be okay?


tfdahlin:
I'm fine with having a contact email forwarded to me like that. Give me a second to look at that file though, I have a couple suggestions

setup.py
```diff
@@ -18,7 +18,7 @@
     name="pytube",
     version=__version__,  # noqa: F821
     author="Nick Ficano, Harold Martin",
```
tfdahlin: Maybe change the author to something like "Pytube organization", so we can avoid changing this as the project team evolves?

setup.py
```diff
@@ -18,7 +18,7 @@
...
-    author_email="nficano@gmail.com, harold.martin@gmail.com",
```
tfdahlin, line 21: If we're going the route of a contact email, we should update this to that.

tfdahlin (changes requested):

RONNCC:
sgtm

setup.py
```diff
@@ -18,19 +18,19 @@
...
-            "pytube = pytube.cli:main", "pytube3 = pytube.cli:main"],},
+            "pytube = pytube.cli:main"],},
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Environment :: Console",
         "Intended Audience :: Developers",
-        "License :: OSI Approved :: MIT License",
+        "License :: OSI Approved :: The Unlicense (Unlicense)",
         "Natural Language :: English",
         "Operating System :: MacOS",
```
RONNCC: might want to add linux

setup.py
nficano: sure, I'll keep this pr open until I get that setup

nficano:
Ready for rereview!

RONNCC:
lgtm - ill leave it to @tfdahlin for his comments

.github/workflows/pypi-publish.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   deploy:
-    if: github.actor == 'nficano' || github.actor == 'RONNCC' || github.actor == 'tfdahlin'
```
tfdahlin: we shouldn't be changing this actually -- this is written specifically so that any of the 3 of us can trigger the pypi deploy, and doesn't have to do with repo name

docs/index.rst
```diff
@@ -8,13 +8,13 @@ Release v\ |version|. (:ref:`Installation<install>`)
   :alt: Pypi
   :target: https://pypi.python.org/pypi/pytube/
 
-.. image:: https://travis-ci.org/nficano/pytube.svg?branch=master
```
tfdahlin, line 11: we should actually get rid of the travis-ci lines since we aren't using it any more

docs/index.rst
```diff
@@ -8,13 +8,13 @@ Release v\ |version|. (:ref:`Installation<install>`)
...
+.. image:: https://travis-ci.org/pytube/pytube.svg?branch=master
    :alt: Build status
-   :target: https://travis-ci.org/nficano/pytube
```
tfdahlin, line 13: same thing here, should get rid of travis-ci

docs/index.rst
```diff
@@ -8,13 +8,13 @@ Release v\ |version|. (:ref:`Installation<install>`)
...
+   :target: https://travis-ci.org/pytube/pytube
 
-.. image:: https://coveralls.io/repos/github/nficano/pytube/badge.svg?branch=master
```
tfdahlin, line 15: pretty sure coveralls isn't around any more either in our CI

setup.py
```diff
@@ -32,11 +32,7 @@
         "Intended Audience :: Developers",
         "License :: OSI Approved :: The Unlicense (Unlicense)",
         "Natural Language :: English",
-        "Operating System :: MacOS",
-        "Operating System :: Microsoft",
-        "Operating System :: POSIX",
-        "Operating System :: Unix",
-        "Operating System :: Linux",
+        "Operating System :: OS Independent",
```
tfdahlin, line 34: oh, good call, i like this better

setup.py
```diff
@@ -48,7 +44,7 @@
         "Topic :: Terminals",
         "Topic :: Utilities",
     ],
-    description=("Python 3 library for downloading YouTube Videos."),
+    description=("Python library for downloading YouTube Videos."),
```
tfdahlin: I think we should keep this as python 3 personally, since we don't support python 2 any more

tfdahlin (changes requested):

.github/workflows/pypi-publish.yml
nficano: erm, artifact of find/replace. good catch

tfdahlin (approved):
Looks good to me!

PR #982, opened 2021-04-28
Remove the chunk size file write limit to speed up downloads.

### LINKED ISSUE #948: [BUG] Download appears to be slowed significantly by default_chunk_size in request.py
The download speed seems to be very slow
It's download for me in speed of 
`4 Megabyte / s (  32mbps )`
Is that normal?

I tried many videos

currently I download video with resolution of 720p

```
os: ubuntu 20.04.2 
python version: Python 3.8.5
pytube version: latest from github - 10.5.1
```
speedtest
```
Hosted by FibreStream (Toronto, ON) [2.19 km]: 3.359 ms
Testing download speed...
Download: 1046.61 Mbit/s
Testing upload speed...
Upload: 1766.06 Mbit/s
```

tfdahlin:
I'm not sure what prompted the decision, but a long time ago, the amount of a file that was written to disk at a time was set to 4kb, while the size of a chunk of video to get downloaded was set to 9MB. This means that for every 9MB of file downloaded, somewhere around 2300 write operations are made to disk, which appears to significantly slow the download

### DIFF
--- a/pytube/request.py
+++ b/pytube/request.py
@@ -12,7 +12,6 @@ from pytube.exceptions import RegexMatchError, MaxRetriesExceeded
 from pytube.helpers import regex_search
 
 logger = logging.getLogger(__name__)
-default_chunk_size = 4096  # 4kb
 default_range_size = 9437184  # 9MB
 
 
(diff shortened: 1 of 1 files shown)


tfdahlin:
@nficano do you know why you would have limited the disk write size to 4kb like this by chance?

Muni-Royal-King:
can any one say how to get file size of an url ... its helpful to download crashed files

tfdahlin:
hi @Muni-Royal-King; it's a little more complicated than that behind the scenes, but resumeable downloads is on the list of things i'm hoping to add at some point in the future.

PR #1018, opened 2021-06-13
support user channels

JarbasAl:
pytube does not seem to support user channels, eg, Vsauce youtube.com/user/Vsauce

this seems to be equivalent to the current Channel class for all intents and purposes, just needed to update the regex that parses urls


tfdahlin:
Looks great, thanks for the contribution! Once the tests pass, I'll merge it

PR #1019, opened 2021-06-13
add support for channel playlists page

JarbasAl:
adds support for parsing channel playlists page

```python
c = Channel(url)
for pl in c.playlists:
    for video in pl.videos:
        print(video.title)
```

I am a bit unsure about pagination, it is not implemented but probably needed

### DIFF
--- a/pytube/contrib/channel.py
+++ b/pytube/contrib/channel.py
@@ -2,10 +2,10 @@
 """Module for interacting with a user's youtube channel."""
 import json
 import logging
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, Iterable
 
 from pytube import extract, Playlist, request
-from pytube.helpers import uniqueify
+from pytube.helpers import uniqueify, cache, DeferredGeneratorList
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tfdahlin:
This code looks good on first inspection, though I would like to test it out myself before I merge it. Unit tests will need to be added before I can merge this, which will mean that we probably need to add some additional mock files. I'm happy to do this sometime later this week, or you can take a stab at it if you'd like to.

pytube/contrib/channel.py
```diff
@@ -199,3 +199,99 @@ def _extract_videos(raw_json: str) -> Tuple[List[str], Optional[str]]:
...
+        try:
+            playlists = initial_data["contents"][
+                "twoColumnBrowseResultsRenderer"][
+                "tabs"][2]["tabRenderer"]["content"][
+                "sectionListRenderer"]["contents"][0][
+                "itemSectionRenderer"]["contents"][0][
+                'shelfRenderer']["content"][
+                'horizontalListRenderer']["items"]
+        except (KeyError, IndexError, TypeError):
+            playlists = []
```
RONNCC, line 226: probably an info warning to explain why it's empty.

pytube/contrib/channel.py
```diff
@@ -199,3 +199,99 @@ def _extract_videos(raw_json: str) -> Tuple[List[str], Optional[str]]:
...
+                    map(
+                        lambda x: (
+                            f"/playlist?list="
+                            f"{x['gridPlaylistRenderer']['playlistId']}"
+                        ),
+                        playlists
+                    )
+                ),
+            ),
+            continuation,
```
RONNCC, line 243: continuation is never populated?

pytube/contrib/channel.py
tfdahlin: No need for an info warning -- this is going to happen on e.g. channels that don't have any playlists created on them. It makes more sense to return an empty array than an error.

pytube/contrib/channel.py
tfdahlin: We don't know how to extract the continuation information currently, because we'd need to find an example channel with something like >50 playlists created on it (which is extremely unlikely to be the case). In the future, when somebody discovers an example of this, we can fill in the continuation code as necessary.

pytube/contrib/channel.py
RONNCC: @tfdahlin  hmm; consideration would be as a user running this on like 40 channels and having some of those be empty --> what would feedback mechanism be to help user debug?

pytube/contrib/channel.py
RONNCC: @tfdahlin  Comedy Central's should have ~ 140+ :-) https://www.youtube.com/c/ComedyCentral/playlists

pytube/contrib/channel.py
tfdahlin: Yeah, we should probably remove the try-catch block honestly

PR #1022, opened 2021-06-18
Fix/1021

tfdahlin:
Stage 1 fix for `get_video_info` url deprecation. This solves the issue for videos that are not age-restricted, but breaks videos that are age restricted.


tfdahlin:
Closing due to age, and changes to pytube that make this code inefficient

PR #1024, opened 2021-06-18
Fix the bug of HTTP Error 404: Not Found

yoonaog01:
add c and cver params in video_info_url and video_info_url_age_restricted


tfdahlin:
Closing in favor of #1025 to use a higher cver

PR #1030, opened 2021-06-23
[DRAFT] Feature/search

### LINKED ISSUE #1001: [FR] Search
It would be cool to have some sort of youtube search (videos/playlists) API in pytube.

tfdahlin:
Adds search functionality via the innertube API. Currently only produces a list of video search results, can later be adapted to include playlist and other results as well.

### DIFF
--- a/docs/user/search.rst
+++ b/docs/user/search.rst
@@ -1,0 +1,50 @@
+.. _search:
+
+Using the search feature
+========================
+
+Pytube includes functionality to search YouTube and return results almost
+identical to those you would find using the search bar on YouTube's website.
+The integration into pytube means that we can directly provide you with
+YouTube objects that can be inspected and dowloaded, instead of needing to do
+additional processing.
+
+Using the Search object is really easy::
+
+    >>> from pytube import Search
+    >>> s = Search('YouTube Rewind')
+    >>> len(s.results)
+    17
+    >>> s.results
+    [\
+        <pytube.__main__.YouTube object: videoId=YbJOTdZBX1g>, \
+        <pytube.__main__.YouTube object: videoId=PKtnafFtfEo>, \
+        ...\
+    ]
+    >>> 
+
+Due to the potential for an endless stream of results, and in order to prevent
+a user from accidentally entering an infinite loop of requesting additional
+results, the ``.results`` attribute will only ever request the first set of
+search results. Additional results can be explicitly requested by using the
+``.get_next_results()`` method, which will append any additional results to
+the ``.results`` attribute::
+
+    >>> s.get_next_results()
+    >>> len(s.results)
+    34
+    >>> 
+
+Additional functionality
+========================
+
+In addition to the basic search functionality which returns YouTube objects,
+searches also have associated autocomplete suggestions. These can be accessed
+as follows::
+
+    >>> s.completion_suggestions
+    [\
+        'can this video get 1 million dislikes', \
+        'youtube rewind 2020 musical', \
+        ...\
+    ]
--- a/pytube/contrib/search.py
+++ b/pytube/contrib/search.py
@@ -1,0 +1,189 @@
+"""Module for interacting with YouTube search."""
+# Native python imports
+import logging
+
+# Local imports
+from pytube import YouTube
+from pytube.innertube import InnerTube
+
+
+logger = logging.getLogger(__name__)
+
+
+class Search:
+    def __init__(self, query):
+        self.query = query
+        self._innertube_client = InnerTube()
+
+        self._results = None
+        self._completion_suggestions = None
+        self._current_continuation = None
+
+    @property
+    def completion_suggestions(self):
+        """Return query autocompletion suggestions for the query.
+
+        :rtype: list
+        :returns:
+            A list of autocomplete suggestions provided by YouTube for the query.
+        """
+        if self._completion_suggestions:
+            return self._completion_suggestions
+        self._completion_suggestions = self.raw_results['refinements']
+        return self._completion_suggestions
+
+    @property
+    def results(self):
+        """Return search results.
+
+        On first call, will generate and return the first set of results.
+        Additional results can be generated using ``.get_next_results()``.
+
+        :rtype: list
+        :returns:
+            A list of YouTube objects.
+        """
+        if self._results:
+            return self._results
+
+        videos, continuation = self.fetch_and_parse()
+        self._results = videos
+        self._current_continuation = continuation
+        return self._results
+
+    def get_next_results(self):
+        """Use the stored continuation string to fetch the next set of results.
+
+        This method does not return the results, but instead updates the results property.
+        """
+        if self._current_continuation:
+            videos, continuation = self.fetch_and_parse(self._current_continuation)
+            self._results.extend(videos)
+            self._current_continuation = continuation
+        else:
+            raise IndexError
+
+    def fetch_and_parse(self, continuation=None):
+        """Fetch from the innertube API and parse the results.
+
+        :param str continuation:
+            Continuation string for fetching results.
+        :rtype: tuple
+        :returns:
+            A tuple of a list of YouTube objects and a continuation string.
+        """
+        # Begin by executing the query and identifying the relevant sections
+        #  of the results
+        raw_results = self.fetch_query(continuation)
+
+        # Initial result is handled by try block, continuations by except block
+        try:
+            sections = raw_results['contents']['twoColumnSearchResultsRenderer'][
+                'primaryContents']['sectionListRenderer']['contents']
+        except KeyError:
+            sections = raw_results['onResponseReceivedCommands'][0][
+                'appendContinuationItemsAction']['continuationItems']
+        item_renderer = None
+        continuation_renderer = None
+        for s in sections:
+            if 'itemSectionRenderer' in s:
+                item_renderer = s['itemSectionRenderer']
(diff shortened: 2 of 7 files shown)


tfdahlin:
@RONNCC care to take a peek?

docs/user/search.rst
```diff
@@ -0,0 +1,50 @@
...
+    >>> 
+
+Due to the potential for an endless stream of results, and in order to prevent
+a user from accidentally entering an infinite loop of requesting additional
+results, the ``.results`` attribute will only ever request the first set of
+search results. Additional results can be explicitly requested by using the
+``.get_next_results()`` method, which will append any additional results to
+the ``.results`` attribute::
+
+    >>> s.get_next_results()
```
RONNCC, line 33: would suggest returning an iterator which has a buffer of $k and therefore obeys the next() interface as possibly more pythonic.

https://wiki.python.org/moin/Iterator

pytube/contrib/search.py
```diff
@@ -0,0 +1,198 @@
...
+from pytube import YouTube
+from pytube.innertube import InnerTube
+
+
+logger = logging.getLogger(__name__)
+
+
+class Search:
+    def __init__(self, query):
+        self.query = query
```
RONNCC, line 20: maybe call it rawQuery as it comes directly from user. Would also suggest python API doc strings if possible similar to other parts of code

RONNCC: since it is input by user

pytube/contrib/search.py
```diff
@@ -0,0 +1,198 @@
...
+        """Use the stored continuation string to fetch the next set of results.
+
+        This method does not return the results, but instead updates the results property.
+        """
+        if self._current_continuation:
+            videos, continuation = self.fetch_and_parse(self._current_continuation)
+            self._results.extend(videos)
+            self._current_continuation = continuation
+        else:
+            raise IndexError
```
RONNCC, line 77: maybe can return StopIteration instead similar to iterator?

pytube/contrib/search.py
```diff
@@ -0,0 +1,198 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Search:
+    def __init__(self, query):
+        self.query = query
+        self._innertube_client = InnerTube()
+
+        self._initial_results = None
+        self._results = None
```
RONNCC: bit confused on different here - it is not immediately apparent

pytube/contrib/search.py
```diff
@@ -0,0 +1,198 @@
...
+                    'url': vid_url,
+                    'title': vid_title,
+                    'channel_name': vid_channel_name,
+                    'channel_url': vid_channel_uri,
+                    'view_count': vid_view_count,
+                    'length': vid_length
+                }
+
+                # Construct YouTube object from metadata and append to results
+                vid = YouTube(vid_metadata['url'])
```
RONNCC, line 188: ideally this would correspond to a new constructor that could possibly take in a kv if needed. but it is minor nit. 
depends on api preferences

RONNCC (commented):
can we add tests as well :-)

docs/user/search.rst
tfdahlin: I like iterators, but I don't think it makes sense in the case of search queries. With e.g. playlists, channel videos, etc., there's a definitive end, so iterators make sense. I don't know for certain if search queries have an end, because youtube likes to be like "hey, maybe you'll want this as well?", and incorporates that into the endpoint.

You could make an argument for using a generator here, since those are designed to support a system for continuously generating output. In fact, I've used a wrapper around generators elsewhere in the codebase to improve our performance. However, I think we should follow more closely with the innertube API's intended use case -- you request a few at first, and if you need more, then you ask for more. Additionally, using generators isn't a great user experience in python. We could do it, but writing something like
```python
s = Search(query)
count = 0
while count < 20:
    curr_vid = next(s.videos)
```
just doesn't really align with how the rest of the library works.

pytube/contrib/search.py
tfdahlin: Yeah, I missed the docstring for __init__, that's my bad. I'll fix that. I don't think raw_query makes much sense since we aren't transforming the user's query behind the scenes, and it only makes other lines longer unnecessarily.

pytube/contrib/search.py
tfdahlin: `StopIteration` only really makes sense if this were implemented as an iterator, see above for my argument against. `IndexError` probably isn't the best either though, might be worth adding a new pytube exception here

pytube/contrib/search.py
tfdahlin: `_initial_results` -- json object returned on first API call. continuation API calls return data that's structured differently, and that does not include some information (notably, search refinement suggestions)
`_results` -- compiled list of YouTube objects.
I'll add some comments to explain the differences.

pytube/contrib/search.py
tfdahlin: What do you mean by that? As in pass the metadata as part of the constructor instead of using setters? Ultimately that's something that I would like to implement, and is why I structure the metadata into a dict even though I don't really use the dict. It's a "planning for future improvements" thing

tfdahlin: Making a new constructor method for `YouTube` objects is outside the scope of this feature, which is why I haven't added it here.

tfdahlin:
> 
> 
> can we add tests as well :-)

I would like to add unit tests, I'm a little hesitant about how to do it here though. I find that most of our unit tests actually aren't as useful as they should be. They often don't really test anything. Good example of this is `test_extract.test_info_url`. Every time that YouTube changes something, I need to go and update various unit tests that have hardcoded values. In the case of the `test_info_url`, I've had to add parameters to the "expected url", and this past time I decided to update it so that I don't need to make those changes as often by making the test less specific in the expected url. Whenever I update these unit tests, I find myself basically running the unit tests and copying it into the test case, which defeats the purpose of the test in the first place.

In the case of this new functionality, I don't really know a good way to make the unit tests useful. If YouTube changes the way they format their API response, then I'm going to be doing more of that "copy and paste to match" again, which means our unit tests... aren't actually validating functionality at all. I don't know a good way around this.

Ideally, the unit tests would *actually* run against YouTube's API, but rate limits mean we get a lot of failures. We also can't make an automated process for updating our mock files for the same reason. I can set up a test suite for running locally where I won't get rate limited, but then it won't be a direct part of CI/CD. Idk the best way forward

docs/user/search.rst
RONNCC: @tfdahlin  would +1 to generator.

> However, I think we should follow more closely with the innertube API's intended use case -- you request a few at first, and if you need more, then you ask for more. Additionally, using generators isn't a great user experience in python. We could do it, but writing something like

Sorry don't think i'm understanding -- we could do the following?

```
buffer = deque()
def __next__( ...):
   if buffer:
       return buffer.pop() 
    if getMore := innerTube.moreResults():
          buffer = getMore[1:]
          return getMore[0]
     return StopIteration
```

pytube/contrib/search.py
RONNCC: makes sense - if you could just add a comment :-) 

my thinking is it would differentiate between user-initiated searches and system-initiated searches (as part of a different feature), but perhaps I've just working too much in midtiers recently

pytube/contrib/search.py
RONNCC: Generator should also return StopIteratioon I believe? https://stackoverflow.com/questions/2322342/should-a-python-generator-raise-an-exception-when-there-are-no-more-elements-to

pytube/contrib/search.py
RONNCC: Ah, interesting distinction - didn't know that

pytube/contrib/search.py
RONNCC: sounds good

RONNCC:
Perhaps we can continue with the using mocks and running against that with 1 'real' query that is used to update the returned mock json?

Definitely agree with difficulty

pytube/contrib/search.py
tfdahlin: We definitely shouldn't be initiating searches ourselves because of network overhead. I also can't imagine a use case where we need to query YouTube's search functionality on behalf of a user where the user wouldn't query it themselves. If we need to provide functionality for that, we can just create new `Search` objects and return those to the user, similar to how I provide `YouTube` objects to the user from search results

pytube/contrib/search.py
tfdahlin: Generators aren't supposed to raise StopIteration themselves (https://www.python.org/dev/peps/pep-0479/). Instead you just return.

tfdahlin:
In the past, when I first wrote some unit tests for a new feature I added, I directly queried against YouTube, but got 429 errors. We cannot reliably get around rate limits, so we can't assume that we can make any "real" queries as part of our test suite, which means we can't update the returned mock json :/ I would love it if we could e.g. make 5-10 requests to YouTube, generate new mocks, and retest the library on a weekly cadence or something, but since we don't control YouTube, we can't get exceptions to rate limiting. 

I'm happy to try and write tests for this, but the tests should mean something imo

pytube/contrib/search.py
RONNCC: Ah oops - i'm a bit outdated there :-)

RONNCC:
sgtm;

Issue #1085, opened 2021-08-06
[BUG] KeyError 'start' when getting captions from a video


gabrielziegler3:
I keep getting `KeyError: 'start'` when I try to get a caption from a video in a playlist.
 
**To Reproduce**
Here is the code I am trying to test:
```python
import pytube
from pytube import Playlist, YouTube

url = "https://www.youtube.com/watch?v=vKA4w2O61Xo&list=PLkahZjV5wKe8WFEwvs69V7JO-Cx57rZ8W"
p = Playlist(url)

for v in p.videos[:3]:
    print("trying to get captions for:", v.title)
    print(v.captions["a.en"].generate_srt_captions())
```

This code used to print the caption before updating pytube, but now it breaks with the following trace:

```python
KeyError                                  Traceback (most recent call last)
~\test_pytube.py in <module>
     10 for v in p.videos[:3]:
     11     print("trying to get captions for:", v.title)
---> 12     print(v.captions["a.en"].generate_srt_captions())

~\AppData\Roaming\Python\Python38\site-packages\pytube\captions.py in generate_srt_captions(s
elf)
     49         recompiles them into the "SubRip Subtitle" format.
     50         """
---> 51         return self.xml_caption_to_srt(self.xml_captions)
     52
     53     @staticmethod

~\AppData\Roaming\Python\Python38\site-packages\pytube\captions.py in xml_caption_to_srt(self
, xml_captions)
     81             except KeyError:
     82                 duration = 0.0
---> 83             start = float(child.attrib["start"])
     84             end = start + duration
     85             sequence_number = i + 1  # convert from 0-indexed to 1.

KeyError: 'start'
```

**System information**
Please provide the following information:
- Python version: Python 3.8.5
- Pytube version: 11.0.0
- Command used to install pytube: pip install -U pytube

gabrielziegler3:
Is there any update regarding this issue? Has the way to download captions changed since version <`11.0.0`?

glubsy:
I have not had a close look at the code, but the start to a possible solution would be to change this:
```python
start = float(child.attrib["start"])
```
to this:
```python
start = float(child.attrib.get("start", 0))
```
This will get rid of the exception, but might produce unwanted consequences. I don't know.

maksimbolonkin:
Apparently YouTube changed their [captions format](https://js-jrod.medium.com/the-first-complete-guide-to-youtube-captions-f886e06f7d9d).

Here's my version of the function for the function in captions.py:

[...]

PR #1086, opened 2021-08-09
close issue: #1085

beancookie:
Resolve the xml format changes, resulting in a parsing srt failure

### DIFF
--- a/pytube/captions.py
+++ b/pytube/captions.py
@@ -72,15 +72,15 @@ class Caption:
             XML formatted caption tracks.
         """
         segments = []
-        root = ElementTree.fromstring(xml_captions)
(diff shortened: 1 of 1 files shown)


JavDomGom (approved):
Although the XML parsing problem is solved, there is still some problem with timing:
```python
yt = YouTube('https://www.youtube.com/watch?v=BGTRqGeIxOA')
captions = yt.captions.get('a.en')
print(captions.generate_srt_captions())
```
Look at the timing of the subtitles, and the text is not extracted either:
```
1
00:00:00,000 --> 00:00:00,000


2
00:00:30,000 --> 01:07:00,000


3
00:26:50,000 --> 01:07:00,000
 

4
00:27:00,000 --> 01:55:30,000
```

PR #1127, opened 2021-09-24
Set bitrate to None when stream does not have bitrate

dreamerlin:
It will raise error if the stream doesn't have 'bitrate'

### DIFF
--- a/pytube/streams.py
+++ b/pytube/streams.py
@@ -57,7 +57,7 @@ class Stream:
         self.video_codec, self.audio_codec = self.parse_codecs()
 
         self.is_otf: bool = stream["is_otf"]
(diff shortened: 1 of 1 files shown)


glubsy:
Wouldn't it make more sense to set it to `None` instead of `0`? Especially since the type hinting is Optional.

glubsy (approved):

rmyosh:
Is there something wrong with this code in its current state? It's been approved for almost a full year now but has yet to be merged. Currently experiencing an issue with older videos which this fixes and I'd like to not have to rely on a feature branch for such a silly bug. I'd also rather not submit an issue for a problem that is already solved with this PR.

PR #1133, opened 2021-10-02
search - detect end of results + iterate_results

JarbasAl:
detects end of results

```
{'messageRenderer': {'text': {'runs': [{'text': 'No more results'}]},
                     'trackingParams': 'CAIQljsYACITCP6uwrXbrPMCFRshBgAdahMLCg=='}}
 ```
adds a method that yields video results as fast as possible and automatically handles pagination

```python
s = Search("rob zombie")
for v in s.iterate_results():
    print(v)
```

also changes log.warn to log.warning, pycharm was complaining about it being deprecated

### DIFF
--- a/pytube/contrib/search.py
+++ b/pytube/contrib/search.py
@@ -149,11 +149,16 @@ class Search:
                 if 'backgroundPromoRenderer' in video_details:
                     continue
 
+                # no more results
+                if 'messageRenderer' in video_details:
+                    next_continuation = None
+                    break
+
                 if 'videoRenderer' not in video_details:
-                    logger.warn('Unexpected renderer encountered.')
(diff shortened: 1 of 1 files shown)


JarbasAl:
I have created a wrapper package so i can use this missing functionality, but would like for this to be integrated directly in pytube. Apologies but I don't currently have the bandwidth to get familiar with pytube in depth and make a top notch PR with unittests and all.

I leave my code here for reference, it subclasses pytube to add new methods https://github.com/OpenJarbas/tutubo

I added "Preview" objects that do not parse the individual search results to make things faster, i also added support for searching more than just videos

Hopefully I can get back to this at some point and submit a proper PR :)

Issue #1163, opened 2021-11-21
[BUG] 'NoneType' object has no attribute 'span'


SuperZombi:
```Python
arr = yt.streams.filter(only_video=True, file_extension='mp4')
```
<details>

```Python
line 292, in streams
    return StreamQuery(self.fmt_streams)
  File "...\Python39\lib\site-packages\pytube\__main__.py", line 177, in fmt_streams
    extract.apply_signature(stream_manifest, self.vid_info, self.js)
  File "...\Python39\lib\site-packages\pytube\extract.py", line 409, in apply_signature
    cipher = Cipher(js=js)
  File "...\Python39\lib\site-packages\pytube\cipher.py", line 44, in __init__
    self.throttling_array = get_throttling_function_array(js)
  File "...\Python39\lib\site-packages\pytube\cipher.py", line 323, in get_throttling_function_array
    str_array = throttling_array_split(array_raw)
```
</details>

```Python
  File "...\Python39\lib\site-packages\pytube\parser.py", line 158, in throttling_array_split
    match_start, match_end = match.span()
AttributeError: 'NoneType' object has no attribute 'span'
```

Pytube version: ```11.0.1```
Python ```3.9.7```

SuperZombi:
Literally yesterday everything worked, but today it gives this error.

BergHeisen:
Got the same just recently

Urgarth:
Same here
```python
yt = YouTube('https://www.youtube.com/watch?v=7BbSo******')
yt.title
'Multistreaming with Restream.io'
yt.thumbnail_url
'https://i.ytimg.com/vi/7Bb****aUyg/sddefault.jpg'
yt.streams
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "C:\Users\Urgarth\PycharmProjects\PyTube\venv\lib\site-packages\pytube\__main__.py", line 292, in streams
    return StreamQuery(self.fmt_streams)
  File "C:\Users\Urgarth\PycharmProjects\PyTube\venv\lib\site-packages\pytube\__main__.py", line 177, in fmt_streams
    extract.apply_signature(stream_manifest, self.vid_info, self.js)
  File "C:\Users\Urgarth\PycharmProjects\PyTube\venv\lib\site-packages\pytube\extract.py", line 409, in apply_signature
    cipher = Cipher(js=js)
[...]

PR #1166, opened 2021-11-22
hotfix parses the throttling array javascript error nontype

thinhpm:
Fix error 'NoneType' object has no attribute 'span'

### DIFF
--- a/pytube/parser.py
+++ b/pytube/parser.py
@@ -152,7 +152,7 @@ def throttling_array_split(js_array):
     func_regex = re.compile(r"function\([^)]+\)")
 
     while len(curr_substring) > 0:
-        if curr_substring.startswith('function'):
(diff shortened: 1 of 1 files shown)


Derpiesaurus:
Not sure if people usually are able to publicly show a like for a pull request as I don't use my Github account often, but I don't see a button so I'll comment that this worked perfectly, so thank you!

89z:
https://github.com/pytube/pytube/issues/1163

glubsy:
Perhaps changing the regex would be beneficial as well, as mentioned [here](https://github.com/pytube/pytube/issues/1163#issuecomment-975428526).

Also adding some tests to the test suite would be a good idea.

RONNCC:
hey @thinhpm  

Thanks for the PR!

PR #1184, opened 2021-12-17
Update streams.py

PeppeSaccardi:
This change does not affect massively the behavior of the class. However prevent crashing due to unavailable streams'bitrate

### DIFF
--- a/pytube/streams.py
+++ b/pytube/streams.py
@@ -57,7 +57,10 @@ class Stream:
         self.video_codec, self.audio_codec = self.parse_codecs()
 
         self.is_otf: bool = stream["is_otf"]
-        self.bitrate: Optional[int] = stream["bitrate"]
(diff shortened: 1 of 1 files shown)


pytube/streams.py
```diff
@@ -57,7 +57,10 @@ def __init__(
         self.video_codec, self.audio_codec = self.parse_codecs()
 
         self.is_otf: bool = stream["is_otf"]
-        self.bitrate: Optional[int] = stream["bitrate"]
+        if "bitrate" in stream.keys():
+            self.bitrate: Optional[int] = stream["bitrate"]
+        else:
+            self.bitrate = 0
```
sacherjj, line 63: I believe using the default get if not found in stream dict will do the same cleaner.

```suggestion
        self.bitrate = stream.get("bitrate", 0)
```

PR #1188, opened 2021-12-27
Fix #1181: Verify Stream attribute before use it.

### LINKED ISSUE #1181: 'Stream' object has no attribute 'fps'
pytube 11.0.2, python 3.10.1

Trying to filter YouTube streams by fps and got  Error.

Input: YouTube("https://www.youtube.com/watch?v=xIANCyXGN5I").streams.filter(fps=30)
Ouput: AttributeError: 'Stream' object has no attribute 'fps'

  File "C:\Users\Main\Desktop\youtubeDL\lib\site-packages\pytube\query.py", line 175, in filter
    return self._filter(filters)
  File "C:\Users\Main\Desktop\youtubeDL\lib\site-packages\pytube\query.py", line 181, in _filter
    return StreamQuery(list(fmt_streams))
  File "C:\Users\Main\Desktop\youtubeDL\lib\site-packages\pytube\query.py", line 129, in <lambda>
    filters.append(lambda s: s.fps == fps)
AttributeError: 'Stream' object has no attribute 'fps'

gaetan1903:
before run filter condition, verify if the attribute exists,

### DIFF
--- a/pytube/query.py
+++ b/pytube/query.py
@@ -123,28 +123,52 @@ class StreamQuery(Sequence):
         """
         filters = []
         if res or resolution:
-            filters.append(lambda s: s.resolution == (res or resolution) if hasattr(s, "resolution") else False)
+            filters.append(
+                lambda s: s.resolution == (res or resolution)
+                if hasattr(s, "resolution") else False
+            )
 
         if fps:
-            filters.append(lambda s: s.fps == fps if hasattr(s, "fps") else False)
+            filters.append(
+                lambda s: s.fps == fps
+                if hasattr(s, "fps") else False
+            )
 
         if mime_type:
-            filters.append(lambda s: s.mime_type == mime_type if hasattr(s, "mime_type") else False)
+            filters.append(
+                lambda s: s.mime_type == mime_type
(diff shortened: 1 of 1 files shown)


gaetan1903:
image

gaetan1903:
image
There is a object doesn't have a fps attribut

gaetan1903:
@glubsy can you check and close this if is bad idea or not compliant?

glubsy (approved):
A few remarks:

* Paste text instead of screenshots. It makes it easier to grab the video ID for others to test, for example:
```python
from pytube import YouTube

yt = YouTube("https://www.youtube.com/watch?v=xIANCyXGN5I")

for k,v in enumerate(yt.streams):
    try:
        print(f"{k}, fps: {v.fps}")    
    except Exception as e:
        print(e)
        print(f"{k}, {v}")

for stream in yt.streams.filter(fps=30):
    print(f"filtered: {stream}")
    assert hasattr(stream, "fps")
```
* It might be a good idea to add some tests in test_query.py

Other than that, after testing a bit, it looks good to me. Good patch!

PR #1194, opened 2022-01-07
Add custom chunk size to request stream function

heindrickdumdum0217:


pytube/cipher.py
```diff
@@ -290,7 +293,8 @@ def get_throttling_function_code(js: str) -> str:
         The name of the function used to compute the throttling parameter.
     """
     # Begin by extracting the correct function name
-    name = re.escape(get_throttling_function_name(js))
+    # name = re.escape(get_throttling_function_name(js))
+    name = "iha"  # temporary fix
```
glubsy: This is a very bad idea, and it will break when you ~least expect it~. edit: actually, you already expect it to break eventually.

glubsy:
I would advise against including unrelated changes in the same pull request. Separate them by creating a feature branch for each change you want to propose.

Issue #1199, opened 2022-01-19
[BUG] RegexMatchError ?


Rizmi:
Until now it's worked perfectly, but now it's showing a issue
is this only for me or ?, can anyone help ty

```
Ignoring exception in on_message
Traceback (most recent call last):
File "/.local/lib/python3.9/site-packages/pytube/__main__.py", line 177, in fmt_streams
extract.apply_signature(stream_manifest, self.vid_info, self.js)
File "/.local/lib/python3.9/site-packages/pytube/extract.py", line 409, in apply_signature
cipher = Cipher(js=js)
File "/.local/lib/python3.9/site-packages/pytube/cipher.py", line 33, in __init__
raise RegexMatchError(
pytube.exceptions.RegexMatchError: __init__: could not find match for ^\\w+\\W
```

AlexanderMelnikov:
Same.
Simple code:
```
from pytube import YouTube  
YouTube('https://www.youtube.com/watch?v=dQw4w9WgXcQ').streams.first().download()
```
Raise error:
```
pytube.exceptions.RegexMatchError: __init__: could not find match for ^\w+\W
```

mishailovic:
lmao, got the same error and wanted to open an issue. Literally yesterday everything was fine

TheHamkerCat:
Prolly some new changes from youtube side

compsage:
ditto

Lolcker2:
got the same error, is it because of the new version?
(the error: pytube.exceptions.RegexMatchError: __init__: could not find match for ^\w+\W)

tylerweston:
As close as I can tell this is being thrown from cipher.py which attempts to thwart YouTubes preventative measures for allowing automated video downloading. If YT wised up and are using a tougher challenge method now this could be a tough nut to crack. Lines 30, 31 in cipher.py
```
        var_regex = re.compile(r"^\w+\W")
        var_match = var_regex.search(self.transform_plan[0])
        if not var_match:
            raise RegexMatchError(
                caller="__init__", pattern=var_regex.pattern
            )
```

Looking through the js that the regex is searching is some giant obfuscated mess that I am in no way qualified to deal with so this is where the extent of my help with this issue ends.

JavDomGom:
Same. Raise exception from here: https://github.com/pytube/pytube/blob/master/pytube/cipher.py#L31

[...]

PR #1200, opened 2022-01-20
Fixed RegexMatchError

DJStompZone:
Fixed `pytube.exceptions.RegexMatchError: __init__: could not find match for ^\w+\W`

Solution proposed by [juanchosaravia](https://github.com/juanchosaravia) here
https://github.com/pytube/pytube/issues/1199#issuecomment-1016783092

### DIFF
--- a/pytube/cipher.py
+++ b/pytube/cipher.py
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 class Cipher:
     def __init__(self, js: str):
         self.transform_plan: List[str] = get_transform_plan(js)
-        var_regex = re.compile(r"^\w+\W")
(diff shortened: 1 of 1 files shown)


boardkeystown:
LETS GOOOOO!!!!! Let's get this pull request done so we can get pytube fixed and I can get it through pip again

HUNTER-28:
fix-regex

lemurmad:
Thanks for sending a fix at a fast pace. I was looking to see if anybody had fixed it.
Thanks again.

RONNCC:
Hey @DJStompZone
 thanks for the PR -- can you add a unit test to show the behavior before and after and then I can merge :-) 

PR #1204, opened 2022-01-22
refactoring code with Assign Multiple Targets

idiomaticrefactoring:
refactoring code with Assign Multiple Targets which is more pythonic, concise, readable and efficient;how do think this change which has practical value？


idiomaticrefactoring:
Is there someone who would like to review my pr?

glubsy (approved):
Looks good.

PR #1215, opened 2022-01-30
easy and fast way to handle dups in list

UnknownCrafts:
Way faster than running a for loop while checking a dictionary to make a duplicate free list, my method also preserves the order of the list.

### DIFF
--- a/pytube/contrib/search.py
+++ b/pytube/contrib/search.py
@@ -39,10 +39,9 @@ class Search:
         :returns:
             A list of autocomplete suggestions provided by YouTube for the query.
         """
-        if self._completion_suggestions:
-            return self._completion_suggestions
         if self.results:
             self._completion_suggestions = self._initial_results['refinements']
+            
         return self._completion_suggestions
 
     @property
--- a/pytube/contrib/search.py
+++ b/pytube/contrib/search.py
@@ -39,6 +39,9 @@ class Search:
         :returns:
             A list of autocomplete suggestions provided by YouTube for the query.
         """
+        if self._completion_suggestions:
+            return self._completion_suggestions
+
         if self.results:
             self._completion_suggestions = self._initial_results['refinements']
             
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -47,10 +47,7 @@ def recording_available(watch_html):
     unavailable_strings = [
         'This live stream recording is not available.'
     ]
-    for string in unavailable_strings:
-        if string in watch_html:
-            return False
-    return True
+    return unavailable_strings.count(watch_html) != 0
 
 
 def is_private(watch_html):
@@ -67,10 +64,8 @@ def is_private(watch_html):
         "\"simpleText\":\"Private video\"",
         "This video is private."
     ]
-    for string in private_strings:
-        if string in watch_html:
-            return True
-    return False
+    
+    return private_strings.count(watch_html) != 0
 
 
 def is_age_restricted(watch_html: str) -> bool:
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -64,7 +64,6 @@ def is_private(watch_html):
         "\"simpleText\":\"Private video\"",
         "This video is private."
     ]
-    
     return private_strings.count(watch_html) != 0
 
 
@@ -84,7 +83,7 @@ def is_age_restricted(watch_html: str) -> bool:
     return True
 
 
-def playability_status(watch_html: str) -> (str, str):
+def playability_status(watch_html: str) -> Tuple(str, str):
     """Return the playability status and status explanation of a video.
 
     For example, a video may have a status of LOGIN_REQUIRED, and an explanation
--- a/pytube/helpers.py
+++ b/pytube/helpers.py
@@ -8,6 +8,7 @@ import re
 import warnings
 from typing import Any, Callable, Dict, List, Optional, TypeVar
 from urllib import request
+from collections import OrderedDict
 
 from pytube.exceptions import RegexMatchError
 
@@ -270,15 +271,8 @@ def uniqueify(duped_list: List) -> List:
     :return List result
         De-duplicated list
     """
-    seen: Dict[Any, bool] = {}
-    result = []
-    for item in duped_list:
-        if item in seen:
-            continue
-        seen[item] = True
-        result.append(item)
-    return result
-
+    return list(OrderedDict.fromkeys(duped_list))
+    
 
 def generate_all_html_json_mocks():
     """Regenerate the video mock json files for all current test videos.
--- a/pytube/helpers.py
+++ b/pytube/helpers.py
@@ -325,6 +325,5 @@ def create_mock_html_json(vid_id) -> Dict[str, Any]:
     logger.info(f'Outputing json.gz file to {gzip_filepath}')
     with gzip.open(gzip_filepath, 'wb') as f:
         f.write(json.dumps(html_data).encode('utf-8'))
-        f.close()
 
     return html_data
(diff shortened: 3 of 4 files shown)


pytube/streams.py
```diff
@@ -271,6 +271,7 @@ def download(
                     bytes_remaining -= len(chunk)
                     # send to the on_progress callback.
                     self.on_progress(chunk, fh, bytes_remaining)
+            fh.close()
```
glubsy: You most likely do not need to call close() on the file handle if you already use a context manager (the `open(...) as fh` above). The context manager calls it automatically for you.

pytube/helpers.py
```diff
@@ -331,5 +325,6 @@ def create_mock_html_json(vid_id) -> Dict[str, Any]:
     logger.info(f'Outputing json.gz file to {gzip_filepath}')
     with gzip.open(gzip_filepath, 'wb') as f:
         f.write(json.dumps(html_data).encode('utf-8'))
+        f.close()
```
glubsy: You most likely do not need to call close() on the file handle if you already use a context manager (the open(...) as f above). The context manager calls it automatically for you.

UnknownCrafts: yea i just realized that "with" already does that. mb

pytube/contrib/search.py
```diff
@@ -39,10 +39,9 @@ def completion_suggestions(self):
         :returns:
             A list of autocomplete suggestions provided by YouTube for the query.
         """
-        if self._completion_suggestions:
-            return self._completion_suggestions
```
glubsy: This does not seem to be a redundant statement. It avoids updating the cache if the value has been set once.

UnknownCrafts: oh

UnknownCrafts:
What do you think about the "easy and fast way to handle dups" commit

glubsy:
> What do you think about the "easy and fast way to handle dups" commit

I see 2 approaches to this question: 1) performance 2) readability

1) I have not benchmarked this, but I doubt this adds much in terms of performance

2) Sure it shortens the function, and became more "readable" in a way, but it also became less intelligible. Someone with less python experience would have to lookup what fromkeys() does and why the OrderedDict is used here.

I think it's alright.

UnknownCrafts:
by "string.count" i meant "list.count"

UnknownCrafts:
btw there is a weird error for me in extract.py
Screen Shot 2022-01-31 at 8 57 31 AM

UnknownCrafts:
not sure if you have it too. I haven't even touched that line

UnknownCrafts:
> > What do you think about the "easy and fast way to handle dups" commit
> 
> I see 2 approaches to this question: 1) performance 2) readability
> 
> 1. I have not benchmarked this, but I doubt this adds much in terms of performance
> 2. Sure it shortens the function, and became more "readable" in a way, but it also became less intelligible. Someone with less python experience would have to lookup what fromkeys() does and why the OrderedDict is used here.
> 
> I think it's alright.

I benchmarked it and found out that the performance difference grows depending on how many times the functions are iterated and not the size of the list.
Screen Shot 2022-01-31 at 9 03 47 AM

UnknownCrafts:
it was a quick benchmark so there might be some other aspects that can still be checked.

glubsy:
Regarding https://github.com/pytube/pytube/pull/1215/commits/fed790557299513d3c27c7f8b662c7c387559bfb, same thing as above, it became a bit less readable, but it's alright in my opinion.

Thanks for taking the time to benchmark your changes.

A few more remarks:

* Better keep separate changes as separate pull requests if possible.
* Post your test code as a gist so that everyone can test on their machine as well. However, make sure to choose more descriptive symbol names. It is a bit difficult to make sense of the variable names at first.
* Regarding the Pylance warning (I don't use Pylance) it's not that important but it is indeed true that it should be Tuple(str, str) instead of (str, str).

UnknownCrafts:
alright i will open other pull requests if I change other stuff

pytube/extract.py
```diff
@@ -67,10 +64,8 @@ def is_private(watch_html):
         "\"simpleText\":\"Private video\"",
         "This video is private."
     ]
-    for string in private_strings:
-        if string in watch_html:
-            return True
-    return False
+
```
glubsy: Just nitpicking, but  it's best to keep the same consistent line spacing style, so I would remove that empty line.

UnknownCrafts: ight

UnknownCrafts:
that would be the last commit for this branch, I will make another branch now

UnknownCrafts:
`from collections import OrderedDict
import time
from typing import Any, Dict


#functions

def useDict(dup_list): 
    return list(OrderedDict.fromkeys(dup_list))

def forLoop(dup_list):
    seen: Dict[Any, bool] = {}
    result = []
    for item in dup_list:
        if item in seen:
            continue
        seen[item] = True
        result.append(item)
    return result

#Simple benchmark
iterations = 1000
_list = [1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1]

#benchmark useDict()
start1 = time.time()
for i in range(iterations): useDict(_list)
end1 = time.time()

#benchmark forLoop()
start2 = time.time()
for i in range(iterations): forLoop(_list)
end2 = time.time()

#get the total time spent in the functions
total_time_useDict = start1 - end1
total_time_forLoop = start2 - end2

#print the result
print("useDict: " + str(total_time_useDict) + " vs " + "forLoop: " + str(total_time_forLoop), "iterations: " + str(iterations))`

UnknownCrafts:
`
from collections import OrderedDict
import time
from typing import Any, Dict


#functions

def useDict(dup_list): 
    return list(OrderedDict.fromkeys(dup_list))

def forLoop(dup_list):
    seen: Dict[Any, bool] = {}
    result = []
    for item in dup_list:
        if item in seen:
            continue
        seen[item] = True
        result.append(item)
    return result

#Simple benchmark
iterations = 1000
_list = [1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1,1, 5, 3, 6, 3, 5, 6, 1]

#benchmark useDict()
start1 = time.time()
for i in range(iterations): useDict(_list)
end1 = time.time()

#benchmark forLoop()
start2 = time.time()
for i in range(iterations): forLoop(_list)
end2 = time.time()

#get the total time spent in the functions
total_time_useDict = start1 - end1
total_time_forLoop = start2 - end2

#print the result
print("useDict: " + str(total_time_useDict) + " vs " + "forLoop: " + str(total_time_forLoop), "iterations: " + str(iterations))
`

UnknownCrafts:
that is the test code i used

glubsy (approved):
Looks good.

UnknownCrafts (commented):
alright

UnknownCrafts:
is someone gonna approve the workflow? I wanna delete this branch soon

RONNCC:
hey @UnknownCrafts 


Thanks for the PR!

Ideally this would 'break' in the future when youtube also changes behavior again and we can show it as a regression.
In addition it would help document the change in the code using tests.

Issue #1214, opened 2022-01-30
Unexpected renderer encountered.


Gabriel-R-Parra:
Pytube Version : 11.0.2
Python Version : 3.9.6
Command used to install : "pip install pytube"
Running on windows 11 current version

**Log output:**
Unexpected renderer encountered.
Renderer name: dict_keys(['showingResultsForRenderer'])
Search term: riot earl sweathshirt

**To reproduce:** 

**1st attempt:** (Received error)

class Streamer:
    def __init__(self, search: str):
        video = Search(search)
        video_one = video[0]
        self.audio = video_one

**Output 1st attempt:**

Unexpected renderer encountered.
Renderer name: dict_keys(['showingResultsForRenderer'])
Search term: riot earl sweathshirt
Please open an issue at https://github.com/pytube/pytube/issues and provide this log output.
<pytube.__main__.YouTube object: videoId=fh7a5hfh7-Y>
______________________________________________________________________
**2nd attempt** (Did not receive error)

def get_vid(search):
    video = Search(search)
    video_one = video.results[0]
    print(video_one)

**Output 2nd attempt**

<pytube.__main__.YouTube object: videoId=fh7a5hfh7-Y>
_____________________________________________________________________

I get the object I want using both of the attempts above, however when the Pytube Search is used in a class  sends an error my way. When used in a function Pytube Search seems to work fine (in my case at least).

UnknownCrafts:
waiting for #1212 merger

PR #1216, opened 2022-02-01
Fix renderer issue

UnknownCrafts:
Possible patch for issue #1214


UnknownCrafts:
why the heck are the other commits here

UnknownCrafts:
Bad pull request

PR #1222, opened 2022-02-01
Fix regex to find throttle function name

glubsy:
The javascript now stores the throttling function name in an array.
cf. fix from yt-dlp at https://github.com/yt-dlp/yt-dlp/commit/48416bc4a8f1d5ff07d5977659cb8ece7640dcd8
Fix https://github.com/pytube/pytube/issues/1218

Context: the throttling function's name used to be a simple variable name, but now it is part of an array of symbols.
The javascript code used to look like this:
```javascript
a.C&&(b=a.get("n"))&&(b=Dea(b),a.set("n",b))}};
```
In the above case, `Dea` is the relevant function name.
Now the javascript code looks like this (white spaces added for clarity):
```javascript
var Bpa = [iha];
a.C && (b = a.get("n")) && (b = Bpa[0](b), a.set("n", b), Bpa.length || iha("")) }};
```
In the above case, `iha` is the relevant function name.
We try to capture the name of the array, as well as the index accessed from this array which should give the function name we are looking for.
If we only have one captured group, it means this is not an array (anymore) and the first capture group is the function name.
If we have 2 captured groups, we assume the first group is an array, and the second group is the index from which the target function name is accessed.

PR #1282, opened 2022-04-15
Fix regex to find throttling function name

kinshuk-h:
This pull request attempts to fix the bug observed in #1281, by generalizing the regex used for finding the throttling function name in the `get_throttling_function_name()` function in `cipher.py`.

Fixes: Issue #1281
Extends: PR #1222

#### **Context**:

The JavaScript code containing the throttling function name from the `base.js` asset file is now of the form:

```js
var $x=[uq]; // 1
a.url="";a.D&&(b=a.get("n"))&&(b=$x[0](b),a.set("n",b),$x.length||uq("")) // 2
```

From the above code fragment, `uq` is the required throttling function name.

The older regex looking for line 2 will not be able to find the same due to a fixed length constraint being imposed on the name of the array variable containing the throttling function.

To detect line 1 and extract the function name, any regex metacharacters such as '$' present in the name of the array variable must be escaped prior to building the next regex.


tests/conftest.py
```diff
@@ -150,13 +150,17 @@ def channel_videos_html():
...
     """
-    file_path = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)),
-        "mocks",
-        "base.js.gz",
-    )
-    with gzip.open(file_path, 'rb') as f:
-        return f.read().decode('utf-8')
+    base_js_files = []
+    for file in ["base.js.gz", "base.js-2022-04-15.gz"]:
```
glubsy: Feel free to rename `base.js.gz` to `base.js-2022-02-04.gz` (in the file system as well). ;)

glubsy (approved):
Looks good to me, seems to work after testing. Thanks for your contribution!

tests/conftest.py
kinshuk-h: Okay 👍

glubsy (approved):

Pythonic-Rainbow (approved):
nice

yeralin:
Are we going to merge it?

glubsy (approved):

RONNCC:
looks good to me  -- thank you @kinshuk-h  :-)

RONNCC:
just pushed to pip also @kinshuk-h  @glubsy

RONNCC:
@glubsy  any idea the config to have github auto push to pip ...

glubsy:
@RONNCC I'm not very familiar with DevOps practices yet, but I believe there is a workflow setup [here](https://github.com/pytube/pytube/blob/master/.github/workflows/pypi-publish.yml) already for that purpose, although it may or may not be actually functional.

I suppose the reason why it didn't push anything to pypi automatically is because the merge you just committed didn't pass the tests (especially flake8 which seems to be set to abort the operation if any mistake is detected). Some minor cleaning up according to flake8 might be needed beforehand.  
Edit: actually the reason is most likely because it only triggers when an actual release it made on Github.
There may also be another problem since the [versions on pypi](https://pypi.org/project/pytube/#history) are not in sync with the version number on Github, and the automated publishing action might fail until versions are in sync again (twine will refuse to overwrite a package version that was already published).

You'll probably find more relevant information in the [python docs](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) (although this document does not use [twine](https://twine.readthedocs.io/en/stable/) contrary to what is setup in pytube currently) or [that random page](https://www.caktusgroup.com/blog/2021/02/11/automating-pypi-releases/) which looks awfully similar to pytube's current github workflow I mentioned above.

RONNCC:
oh haha so it is - i'll have to follow up on that later; appreciate it @glubsy  as well as your assistance more generally! 

If you have time would love your assistance cleaning up things around the repo (issues/PRs) etc

glubsy:
@RONNCC I can try helping as much as I can.

PR #1292, opened 2022-04-19
Line number 288 --> nfunc=re.escape(function_match.group(1))),

AmmarFarid10:
Can we please update the line as it is giving an error Pytube Error: get_throttling_function_name: could not find match for multiple.


glubsy:
Duplicate of #1282

PR #1336, opened 2022-06-05
fixed regex match error

brilliant-ember:
Fixes this bug https://github.com/pytube/pytube/issues/1199

Essentially the fix is just updating the regular expression. Tested locally and download is working after the fix on Ubuntu 20.04.

### DIFF
--- a/pytube/cipher.py
+++ b/pytube/cipher.py
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 class Cipher:
     def __init__(self, js: str):
         self.transform_plan: List[str] = get_transform_plan(js)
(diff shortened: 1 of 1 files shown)


1hachem:
why wasn't this merged ?

BuildWithData:
@1hachem seems like I'm not the only one who ended up here today... just tried to download a video and out of no where pytube stopped working. Did you ? Any idea why this patch was not needed yesterday ?

1hachem:
No idea why it stopped working, but it seems that this keeps happening from time to time, it works that it doesn't than it works again, and this is why they didn't patch it

Issue #1258, opened 2022-02-25
[BUG] KeyError: 'content-length'


NishantPacharne:
**The bug**
Iam getting the KeyError: 'content-length' while downloading videos, specifically for this one, when ever I try to download this video in 1080p, I get the error.

**To Reproduce**
Try downloading this video in 1080p resolution with pytube.
https://www.youtube.com/watch?v=6PDxyCLaTeU

**Code**
```
from pytube import YouTube
yt = YouTube("https://www.youtube.com/watch?v=6PDxyCLaTeU")

yt.streams.filter(resolution="1080p").first().download()
```

**Exception**
return int(head(url)["content-length"])
KeyError: 'content-length'

**System information**
Please provide the following information:
- Python version : Python 3.9.2
- Pytube version : 12.0.0
- Command used to install pytube : pip install pytube

tiff1002:
I am having the same issue with the following video:
https://www.youtube.com/watch?v=UQ5_ehVI7mg

the following commands trigger the exception:

`
py = self.pytube_obj.streams.order_by('resolution')
py.desc().first().download(filename_prefix="video_")
`

exception:

`
File "C:\Users\tiff\AppData\Roaming\Python\Python310\site-packages\pytube\streams.py", line 247, in download
bytes_remaining = self.filesize
File "C:\Users\tiff\AppData\Roaming\Python\Python310\site-packages\pytube\streams.py", line 146, in filesize 
self._filesize = request.filesize(self.url)
File "C:\Users\tiff\AppData\Roaming\Python\Python310\site-packages\pytube\request.py", line 200, in filesize
return int(head(url)["content-length"])
KeyError: 'content-length'
`

this is so frustrating

cybermajestic-zz:
Did you resolve this bug?

rishabh3354:
Be smart, move to some other lib, this lib is dead,
I moved my project to youtubedlp

US366:
I'm a beginner so I don't know the details, but ...

I get an error with the type of video codec.

In my environment
download OK
avc1.4d401e
        64002a
        640028
        64001f
        4d4015

NG
avc1.64001e
        4d4016
        4d4014
        4d400c
        4d400b

I'm glad if you can use it as a reference

DaliaO15:
I have the same problem with the video  https://www.youtube.com/watch?v=OYS2pt9DVsY when downloading with tag 136.

Issue #912, opened 2021-01-11
[BUG] KeyError: 'content-length'


DaFluffyPotato:
**Describe the bug**
The download function in 10.4.1 results in a `'content-length'` KeyError.

**To Reproduce**
Here's my code:
```python
from pytube import YouTube

url = input('URL: ')

yt = YouTube(url)
yt.streams.filter(file_extension='mp4').first().download()
```

**Expected behavior**
I expect it to download the video.

**Output**
```
URL: https://www.youtube.com/watch?v=EbNr015DZbM
Traceback (most recent call last):
  File "C:\Users\cmlsc\Desktop\youtube_download.py", line 7, in <module>
    yt.streams.filter(file_extension='mp4').first().download()
  File "D:\Python38\lib\site-packages\pytube\streams.py", line 251, in download
    bytes_remaining = self.filesize
  File "D:\Python38\lib\site-packages\pytube\streams.py", line 158, in filesize
    self._filesize = request.filesize(self.url)
  File "D:\Python38\lib\site-packages\pytube\request.py", line 127, in filesize
    return int(head(url)["content-length"])
KeyError: 'content-length'
```

**System information**
Please provide the following information:
Python Version: 3.8.6
Pytube Version: 10.4.1
Install Command: `D:\Python38\python.exe -m pip install git+https://github.com/pytube/pytube --upgrade`

tfdahlin:
hi @DaFluffyPotato I'm unable to replicate this problem. Can you run it again and see if you have the same issue?

DaFluffyPotato:
I just tested it again. It appears to be working on my end now. I suspect it may be an issue with livestreams archives from livestreams that were recently finished. I'm pretty sure YouTube does some processing for a few hours (or maybe a couple days?) on livestreams after they finish since you can't see the live chat immediately. I'll close this issue and someone can just make a new one if it turns out that there's an issue with fresh livestream archives.

tfdahlin:
sounds good!

dovewing:
i have a same issue
`~\anaconda3\lib\site-packages\pytube\request.py in filesize(url)
    198     :returns: int: size in bytes of remote file
    199     """
--> 200     return int(head(url)["content-length"])
    201 
    202 

KeyError: 'content-length'`

Robin-Haupt-1:
I am also getting the same issue only with the following video, which isn't a recent video or a livestream:

https://www.youtube.com/watch?v=qT4a73ctcmw

```
[...]

Issue #160, opened 2017-10-12
Content Length Bug


emmanuelbernieri:
Hello, 

When I try to execute the example code on a python 2.7 on LINUX I get an error: 

>>> from pytube import YouTube
>>> YouTube('https://youtu.be/9bZkp7q19f0').streams.first().download()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pytube/streams.py", line 161, in download
    bytes_remaining = self.filesize
  File "/usr/local/lib/python2.7/dist-packages/pytube/streams.py", line 138, in filesize
    return int(headers['Content-Length'])
KeyError: 'Content-Length'

Can you help me ?

nficano:
Hi, @emmanuelbernieri I just pushed a patch release that should eliminate this bug. just run:

```bash
pip install -U pytube
```

sorry you're having an issue, and thanks for reporting it! let me know if that doesn't resolve it.

emmanuelbernieri:
it's working now :) I'll let you know if I see new bugs

DiegoGuisasola:
Hello everyone. Apparently is came back:

![image](https://user-images.githubusercontent.com/44145128/105574263-596c9100-5d31-11eb-94a4-ccc7ae9d9eae.png)

Any work around? Thanks!

pgshepherd:
This happens when the stream was a Live Stream, and therefore there is no content-length associated with the stream.  Some have attempted workarounds that avoid the need to know how long the stream is by downloading it in chunks:
https://stackoverflow.com/questions/55631634/recording-youtube-live-stream-to-file-in-python

PR #1345, opened 2022-06-25
Fix content-length #1258, #912, #160

Justxd22:
Content-length error is caused by YouTube not sending content-length when a head request is made asking for media headers before download, 
content-length is used to get download size to be displayed in progress bar and progress hooks so it won't affect our download if it's incorrect, in this case using approximated size is fine. (better than nothing)
 
Possible fix of  #1258, #912, #160

Some of the issues was closed without a solution

mp4 & without audio streams only have the error

Here's some links to test:

https://youtu.be/xrXsm5YgiiI      #all streams except streams with audio
https://youtu.be/UQ5_ehVI7mg  #itag 160
https://youtu.be/6PDxyCLaTeU  #itag 160 and sometimes 360p&1080p
https://youtu.be/qT4a73ctcmw #itag 160

### DIFF
--- a/pytube/streams.py
+++ b/pytube/streams.py
@@ -151,7 +151,7 @@ class Stream:
             except Exception as e:
                 if "content-length" in str(e):
                     self._filesize = int((self._monostate.duration * self.bitrate) / 8)
-                else: print(e)
+                else: raise(e)
         return self._filesize
 
     @property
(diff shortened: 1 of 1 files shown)


ArtiiP:
`         except Exception as e:`

it's  generaly bad idea to catch all exeptions  
why not KeyError only?

edit: also its better reraise uknown exceptions `else: raise e `

Justxd22:
@ArtiiP because content-length error is just a string (content-length isn't a KeyError), i would have to define a new exception to make it work,

you're right it's better to `raise(e)`

wjkoh:
I got the same error when I tried to download the MP4 audio of https://www.youtube.com/watch?v=6QySKWkYlPc.

PR #1364, opened 2022-08-03
Introducing filesize returned in kb, mb and gb

s1ddly:
Adding in properties for the filesize in KB, MB and GB.
These are returned as integers rounded to the nearest significant figure(this because the original filesize attribute is an integer as well and I wanted to continue with that trend)
I have only put in kb, mb and gb, I don't foresee TB large videos on youtube for a while...

The reason I added this in is because I was going to implement this in my own code on top of the pytube library, but I figured it might be better to add it in for everyone else who might use this code base too.

The 2 things I have doubts on are the rounding functionality and integer return type.
if you have set yourself a 1 gigabyte file limit, but the file is 1.45 gb, stream.filesizegb will return 1gb, but a 550mb file will do the same too.
The rounding to return an integer value because the original filesize property returns an int, although for gigabyte large videos, this could mean a +-500mb misrepresentation of the actual file size.  

Happy to discuss with you on the above and find the best solution for this.

### DIFF
--- a/pytube/streams.py
+++ b/pytube/streams.py
@@ -8,6 +8,8 @@ separately).
 """
 import logging
 import os
+from math import ceil
+
 from datetime import datetime
 from typing import BinaryIO, Dict, Optional, Tuple
 from urllib.error import HTTPError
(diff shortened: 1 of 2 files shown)


s1ddly:
After some though I found a way forward on the 2 points I had.
1. Rounding to 3 sigfigs, because this level of fidelity allows kb level definition even at the mb level, and mb level for gb
2. Return type to be float, I chose this because there will be no backwards compatibility issues as these are new features
I chose to have it round up as well, because I would much rather be a bit below than a bit over when it comes to file size.

pytube/streams.py
```diff
@@ -61,6 +63,15 @@ def __init__(
 
         # filesize in bytes
         self._filesize: Optional[int] = int(stream.get('contentLength', 0))
+        
+        # filesize in kilobytes
+        self._filesizekb: Optional[float] = float(ceil(stream.get('contentLength', 0) / 1024 * 1000) / 1000)
```
RONNCC:
```suggestion
        self._filesize_kb: Optional[float] = float(ceil(stream.get('contentLength', 0) / 1024 * 1000) / 1000)
```
could we add another _ here to delimit

RONNCC (commented):
appreciate the PR!

minor nit to rename the vars/ function to use _ via https://peps.python.org/pep-0008/#function-and-variable-names


but otherwise lgtm - feel free to reping on update and i can merge

@s1ddly

pytube/streams.py
RONNCC: and use _ more generally to delimit mb/kb/gb in the function names and variables corresponding to the new functionality

(style via https://peps.python.org/pep-0008/#function-and-variable-names)

s1ddly:
Hi @RONNCC, 
Very valid point you have raised and I'll keep this in mind as I continue to contribute to new projects :)
I've updated the code with your points and also fixed up the testing script.
Feel free to merge when you are ready!

RONNCC (approved):

RONNCC:
Thanks @s1ddly appreciate the contribution here! :-)
Merged ... forget if we built out the autoupdate pipeline for pip yet ...

glubsy:
A bug (#1389) has been reported related to this merge.

The 'contentLength' property is not guaranteed to be an int, you need to introduce checks or at the very least cast the value to int.

PR #1386, opened 2022-09-04
fix captions.generate_srt_captions()

tonydavis629:
Seems that recently Youtube changed how it handles captions, and now captions.generate_srt_captions() is broken:

```
yt = YouTube('https://www.youtube.com/watch?v=gCMGFZ2A85o')

yc = yt.captions['en']
a = yc.generate_srt_captions()

  File "/home/tony/github/videorhyme/videorhyme/youtube_dl/test.py", line 6, in <module>
    a = yc.generate_srt_captions()
  File "/home/tony/miniconda3/envs/videorhyme/lib/python3.10/site-packages/pytube/captions.py", line 51, in generate_srt_captions
    return self.xml_caption_to_srt(self.xml_captions)
  File "/home/tony/miniconda3/envs/videorhyme/lib/python3.10/site-packages/pytube/captions.py", line 83, in xml_caption_to_srt
    start = float(child.attrib["start"])
KeyError: 'start'
```

This PR fixes that problem. I took this code from a year old [post](https://stackoverflow.com/questions/68780808/xml-to-srt-conversion-not-working-after-installing-pytube) on stack overflow, cleaned it up, and it seems to work.

### DIFF
--- a/pytube/captions.py
+++ b/pytube/captions.py
@@ -65,6 +65,7 @@ class Caption:
         ms = f"{fraction:.3f}".replace("0.", "")
         return time_fmt + ms
 
+
     def xml_caption_to_srt(self, xml_captions: str) -> str:
         """Convert xml caption tracks to "SubRip Subtitle (srt)".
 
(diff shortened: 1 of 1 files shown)


hp-luo:
so  what is the update version of pytube? I use the 12.1.0 version ,but it is also the error

tonydavis629:
You can replace the captions.py code with this code to fix it.

Issue #1389, opened 2022-09-17
[BUG] unsupported operand type(s) for /: 'str' and 'int'


whale9820:
I have already installed the latest version of PyTube from Github.

This issue occurs while this line of code is being executed:
`video = yt.streams.filter(only_audio=True).first()`
The `yt` has already been previously defined as `YouTube("https://youtu.be/dQw4w9WgXcQ")`.

Below is the full traceback:
<img width="518" alt="Screenshot 2022-09-17 at 10 10 19 AM" src="https://user-images.githubusercontent.com/87256750/190836460-615e84c5-1507-4fd2-9177-b489ce4cbb30.png">

glubsy:
Looks like this bug has been introduced by #1364 @s1ddly

s1ddly:
Hi @glubsy and @walesworksltd, thanks for brining this to my attention!
I have put in a bug fix here: https://github.com/pytube/pytube/pull/1390
Turns out it was an issue with the type coercion for stream.get('contentLength', 0) as its returning a string instead of a float.

RONNCC:
fix should have been merged :-)

PR #1390, opened 2022-09-17
Bug fix for issue #1389

s1ddly:
[walesworksltd](https://github.com/walesworksltd) has found a bug with the latest code in the commit I have pushed through...
This is shown here in [issue #1389](https://github.com/pytube/pytube/issues/1389).

I am adding in this commit to fix that bug and also further reinforce the _ naming convention on some of the function calls.


s1ddly:
Hi @RONNCC, a bug has been found in this code here: https://github.com/pytube/pytube/issues/1389
I have put the fix in this PR, feel free to have a look when you get a chance.

PR #1413, opened 2022-11-05
Update query.py

ghost:
Add option to get highest resolution for adaptive streams.


RONNCC:
believe this should be fixed in recent changes? Please reopen PR if that's not the case :-)

PR #1419, opened 2022-11-14
certificate verify failed: unable to get local issuer certificate

Asada1:
Hi theres is an issue with working with pytube on macOS. I want to suggest to add information about it and how to solve it in read.me file.
https://stackoverflow.com/questions/52805115/certificate-verify-failed-unable-to-get-local-issuer-certificate


RONNCC:
believe this should be fixed in recent changes? Please reopen PR if that's not the case :-)

PR #1441, opened 2022-12-29
Add missing import statement

ethomas2:
`json_captions()` function makes a call to `json` but doesn't import `json`


rgryta (approved):

RONNCC:
believe this should now be fixed?

PR #1447, opened 2023-01-03
fixed Render issue caused by shorts recommendation

ShahrukhAbrar:
Render issue caused by shorts recommendation in searches.


RONNCC:
is this still an issue?

PR #1471, opened 2023-02-04
Updated According To YouTube's new updated & solved some error

ghost:
## Updated `setup.py`
- Added [`bs4`](https://github.com/wention/beautifulSoup4/) as dependency.

## Updated `channel.py`
- It can now find channel by YouTube's new channel links with `@`.
    - Example: `https://youtube.com/@Youtube`
- It can also give channel description text.

## Updated `search.py`
- It will not show this mess:
>  ```py
> Unexpected renderer encountered.
> Renderer name: dict_keys(['showingResultsForRenderer'])
> Search term: TCAGH2226570
> Please open an issue at https://github.com/pytube/pytube/issues and provide this log output.
> ```

### DIFF
--- a/pytube/contrib/channel.py
+++ b/pytube/contrib/channel.py
@@ -2,8 +2,9 @@
 """Module for interacting with a user's youtube channel."""
 import json
 import logging
+import requests
 from typing import Dict, List, Optional, Tuple
-
+from bs4 import BeautifulSoup
 from pytube import extract, Playlist, request
 from pytube.helpers import uniqueify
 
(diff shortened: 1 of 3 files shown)


pytube/contrib/channel.py
```diff
@@ -21,6 +22,9 @@ def __init__(self, url: str, proxies: Optional[Dict[str, str]] = None):
         """
         super().__init__(url, proxies)
 
+        if "/@" in url:
+            url = BeautifulSoup(requests.get(url).text, "html.parser").find_all("meta", property="og:url")[0].get("content")
+
```
glubsy, line 27: Making a request in the constructor is really not ideal.

glubsy (changes requested):
I don't think we need beautifulsoup here. This is overkill and should probably be achieved with a simple regex.

PR #1496, opened 2023-03-10
Minor documentation fix (add missing import to example)

samryan18:
Need to import `VideoUnavailable` exception to [this example](https://pytube.io/en/latest/user/exceptions.html).


RONNCC:
Makes sense - thanks for the contribution! :-)

PR #1516, opened 2023-03-18
FIX 'NoneType' object has no attribute 'span'

mr-ZIAD:
credit: @dark9ive


eyalBr7:
some repo manager merge this

PR #1518, opened 2023-03-19
Temporary fix Pytube - cipher.py

MikeCVermeer:
Something on Youtube's end broke Pytube. My temporary solution involves bypassing the code that was causing the error by setting the transform plan raw to the javascript code itself, effectively disabling the code that was causing the error. This will get Pytube working again as of 19/03/2023


AlexanderGermek (approved):

entrophae (approved):

RONNCC:
is this still an issue?

RONNCC:
assuming not an issue any longer - please reopen if that's not true 😄

PR #1520, opened 2023-03-19
Fix parser py

fischcode:
I managed to figure out why pytube suddenly dumped: The function find_object_from_startpoint in pytube parser.py failed to recognize regular expressions and ignore closing characters contained therein, resulting in incomplete code fragments being passed to downstream functions.
This fix enables find_object_from_startpoint to recognize and treat regular expressions correctly and restores the original cipher.py.
I have tested that the function find_object_from_startpoint works and returns complete functions, but have no idea how I could test if the mechanism to avoid throttling still works.

### DIFF
--- a/pytube/parser.py
+++ b/pytube/parser.py
@@ -134,12 +134,11 @@ def parse_for_object_from_startpoint(html, start_point):
     full_obj = find_object_from_startpoint(html, start_point)
     try:
         return json.loads(full_obj)
-    except Exception as e:
+    except json.decoder.JSONDecodeError:
         try:
             return ast.literal_eval(full_obj)
         except (ValueError, SyntaxError):
             raise HTMLParseError('Could not parse object.')
-
 
 
 def throttling_array_split(js_array):
--- a/pytube/parser.py
+++ b/pytube/parser.py
@@ -74,18 +74,23 @@ def find_object_from_startpoint(html, start_point):
 
     # First letter MUST be a open brace, so we put that in the stack,
     # and skip the first character.
+    last_char = '{'
+    curr_char = None
     stack = [html[0]]
     i = 1
 
     context_closers = {
         '{': '}',
         '[': ']',
-        '"': '"'
+        '"': '"',
+        '/': '/' # javascript regex
     }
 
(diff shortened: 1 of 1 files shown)


ghost:
@fischcode My solution turned out to be slightly more complex: sluggish-yard/pytube-saguaro#2. I would appreciate if you take a look to see if there is something missing.

fischcode:
You may want to have a look at the discussion in https://stackoverflow.com/questions/5519596/when-parsing-javascript-what-determines-the-meaning-of-a-slash : Eventually, I opted for the simplest solution judged to be suitable for a lexer in most real-world situations: 'JSLint's method of checking the preceding character for (,=:[!&|?{}; is probably sufficient'.
If I understand your code correctly, it makes a number of assumptions all of which must not always be true:
- regular expressions are always preceded by a comma 
- the comma is separated from the first slash by no more than three characters of white space
- regular expressions do not contained escaped slashes \/
- regular expressions do not exceed 20 characters in length
While my solution does not cover all the exotic situations lined out the discussion, it allows for many more preceding separating tokens and makes none of the other assumptions.
Happy coding!

fischcode:
You may want to have a look at the discussion in https://stackoverflow.com/questions/5519596/when-parsing-javascript-what-determines-the-meaning-of-a-slash : Eventually, I opted for the simplest solution judged to be suitable for a lexer in most real-world situations: 'JSLint's method of checking the preceding character for (,=:[!&|?{}; is probably sufficient'.
If I understand your code correctly, it makes a number of assumptions all of which must not always be true:


- regular expressions are always preceded by a comma
- the comma is separated from the first slash by no more than three characters of white space
- regular expressions do not contained escaped slashes \\/
- regular expressions do not exceed 20 characters in length

While my solution does not cover all the exotic situations lined out the discussion, it allows for many more preceding separating tokens and makes none of the other assumptions.

Happy coding!

ghost:
@fischcode Thanks, I've [adopted](https://github.com/sluggish-yard/pytube-saguaro/pull/5) your version and added a fix for mix strings(`'"'`, `"'"`).

PR #1523, opened 2023-03-19
Update search

barisceliker1:
When we search some keywords ,we get error message "unexpected renderer encountered" .Because of that we should skip short videos.


sluggish-yard:
Is there a way to reproduce the issue?

rgryta:
Why skip Shorts instead of taking them in additionally to normal videos? These are valid search results, not "recommended".
This PR is also a duplicate of multiple others though.

PR #1524, opened 2023-03-19
Update Search

barisceliker1:
I just fixed reelShelfRenderer method. Right now there are not error. it counts also youtube short videos.


nficano:
This PR fails the unit test. Once that's fixed I'll merge right away

barisceliker1:
How can I fix this problem ? What about this failing ?

PR #1534, opened 2023-04-08
fix logger.warn deprecation

bvandrc:
Fix logger.warn deprecation. Is causing excess warnings in the console:

image


RONNCC:
makes sense -- thanks

PR #1536, opened 2023-04-10
fix: fix 'json' is not defined

### LINKED ISSUE #1535: [NameError]  'json' is not defined
## **BUG**
When I try to get the `json_captions` from the `pytube.captions.Caption` class, error occurs.

## The following code is causing the bug:
```python
from pytube import YouTube
from pytube.captions import Caption
from pytube.query import CaptionQuery

yt = YouTube("https://youtu.be/8JJ101D3knE")

# captions
all_captions: CaptionQuery = yt.captions
eng_captions: Caption = yt.captions["en"]

# json captions
json_caption = eng_captions.json_captions
```

## **Expected behavior**
A python `dict` of caption tracks.

## **Output**
```sh
parsed = json.loads(text)
NameError: name 'json' is not defined
```

## **System information**
- Python version: `Python 3.10.8`
- Pytube version: `pytube 12.1.3`
- Command used to install pytube: `pip install pytube`

subhranil2605:
This fixes the error occuring when try to get the JSON Caption.

BREAKING CHANGE
Before this fixes, NameError occurs, and can't get the json_caption

Closes #1535

test: add test case for test_captions

This test case adds the testing of the json_caption() function. Add the get_url_for_json() function for this purpose.

### DIFF
--- a/pytube/captions.py
+++ b/pytube/captions.py
@@ -4,6 +4,7 @@ import time
 import xml.etree.ElementTree as ElementTree
 from html import unescape
 from typing import Dict, Optional
+import json
 
 from pytube import request
 from pytube.helpers import safe_filename, target_directory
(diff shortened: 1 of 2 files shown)


RONNCC:
do you have an example of the error by chance?

pytube/captions.py
```diff
@@ -45,12 +46,16 @@ def xml_captions(self) -> str:
     @property
     def json_captions(self) -> dict:
         """Download and parse the json caption tracks."""
-        json_captions_url = self.url.replace('fmt=srv3','fmt=json3')
+        json_captions_url = self.get_url_for_json()
```
RONNCC, line 49: i'm unsure of the need for this change ... seems one line

pytube/captions.py
```diff
@@ -4,6 +4,7 @@
 import xml.etree.ElementTree as ElementTree
 from html import unescape
 from typing import Dict, Optional
+import json
```
RONNCC, line 7: makes sense .. happy to add this line

pytube/captions.py
subhranil2605: Yes, I have created another function, as I saw that there is no Test code for the function, I wrote a test code. For this purpose I have to make another function. Please review that

RONNCC:
should be fixed now I believe

PR #1538, opened 2023-04-11
Update captions.py

tsaijamey:
修复了captions中xml_caption_to_srt方法中对于xml文档结构的错误解析方法。详细说明与示例见代码内的注释文档。

### DIFF
--- a/pytube/captions.py
+++ b/pytube/captions.py
@@ -79,26 +79,68 @@ class Caption:
 
         :param str xml_captions:
             XML formatted caption tracks.
+
+
+        Bug fixed @2023-04-11 by tsaijamey@github :
+        原代码运行时报错  KeyError 'start'，所以根据最新的xml进行了修正。
+        xml的结构已经发生了变化，样例如下：
+        ```
+            <body>
+                <w t="0" id="1" wp="1" ws="1"/>
+                    <p t="60" d="6180" w="1">
+                        <s ac="248">I</s>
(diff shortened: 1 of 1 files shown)


choies1:
Please update `xml_caption_to_srt()` function to work.

The following code works well now.

```python
def xml_caption_to_srt(self, xml_captions: str) -> str:
    """Convert xml caption tracks to "SubRip Subtitle (srt)".

    :param str xml_captions:
        XML formatted caption tracks.
    """
    segments = []
    root = ElementTree.fromstring(xml_captions)
    for i, child in enumerate(list(root.findall('body/p'))):
        text = ''.join(child.itertext()).strip()
        if not text:
            continue
        caption = unescape(text.replace("\n", " ").replace("  ", " "),)
        try:
            duration = float(child.attrib['d'])
        except KeyError:
            duration = 0.0
        start = float(child.attrib['t']) / 1000.0
        end = start + duration / 1000.0
        sequence_number = i + 1  # convert from 0-indexed to 1.
        line = "{seq}\n{start} --> {end}\n{text}\n".format(
            seq=sequence_number,
            start=self.float_to_srt_time_format(start),
            end=self.float_to_srt_time_format(end),
            text=caption,
        )
        segments.append(line)
    return "\n".join(segments).strip()
```

And please add the following line to import parts.

```python
import json
```

RONNCC:
updated in recent commit

Issue #1558, opened 2023-04-27
[BUG] KeyError when downloading a video and/or getting infos about the video


flloschy:
**Describe the bug**
I was redoing a script which automatically downloads YouTube videos. \
In the code I test for the length of the video to not download videos under three minutes. When testing I got the error found below. \
After some trying around I tested the example code which I found in the [docs](https://pytube.io/en/latest/) and got an similar error having something to do with the `vid_info`. \
Yesterday my old script didnt work (which is why im recoding it) because the decapi-youtube API had some problems, I think its possible that this is somewhat the same reason here. \
(decicus *(the maker of decapi)* said "YouTube's API seems to be having issues at the moment, so it may not consistently work." after the API went down, maybe they changed something)

**To Reproduce**
(this is the code found in the [docs](https://pytube.io/en/latest/))
- Open console and start python
- import pytube.Youtube by `>>> from pytube import YouTube`
- try downloading a video by `>>> YouTube('https://youtu.be/9bZkp7q19f0').streams.first().download()`
- get an error

**Expected behavior**
Download the video and not throw an error

**Output**
Error when using `YouTube('https://youtu.be/9bZkp7q19f0').streams.first().download()`
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\python\Lib\site-packages\pytube\__main__.py", line 296, in streams
    return StreamQuery(self.fmt_streams)
                       ^^^^^^^^^^^^^^^^
  File "D:\python\Lib\site-packages\pytube\__main__.py", line 176, in fmt_streams
    stream_manifest = extract.apply_descrambler(self.streaming_data)
                                                ^^^^^^^^^^^^^^^^^^^
  File "D:\python\Lib\site-packages\pytube\__main__.py", line 161, in streaming_data
    return self.vid_info['streamingData']
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'streamingData'
```

Errors when using `YouTube('https://youtu.be/9bZkp7q19f0').title`
```
Traceback (most recent call last):
  File "D:\python\Lib\site-packages\pytube\__main__.py", line 341, in title
    self._title = self.vid_info['videoDetails']['title']
                  ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'videoDetails'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

Issue #1557, opened 2023-04-27
[BUG]KeyError: 'videoDetails' ; Exception while accessing title of https://youtube.com/watch?v=AVYGCjwjUXg. Please file a bug report at https://github.com/pytube/pytube


kevinb-dev:
My code :
```from pytube import YouTube

url = "https://www.youtube.com/watch?v=AVYGCjwjUXg&t=88s"


def on_download_progress(stream, chunk, bytes_remaining):
    bytes_downloaded = stream.filesize - bytes_remaining
    percent = bytes_downloaded * 100 / stream.filesize
    print(f"Progression du téléchargement : {int(percent)}%")


youtube_video = YouTube(url)

youtube_video.register_on_progress_callback(on_download_progress)

print(f"Titre: {youtube_video.title}")
print(f"Nombre de vues : {youtube_video.views}")

print("Liste des streams")
for stream in youtube_video.fmt_streams:
    print(f" {stream}")

# stream = youtube_video.streams.get_by_itag(18)
stream = youtube_video.streams.get_highest_resolution()
print("Téléchargement ...")
stream.download()
print(f"Téléchargement effectué, la vidéo a une résolution de {stream.resolution}")
```

And the error : 
Traceback (most recent call last):
  File "C:\Users\Skoll\PycharmProjects\YouTube Downloader\venv\lib\site-packages\pytube\__main__.py", line 341, in title
    self._title = self.vid_info['videoDetails']['title']
KeyError: 'videoDetails'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Skoll\PycharmProjects\YouTube Downloader\main.py", line 20, in <module>
    print(f"Titre: {youtube_video.title}")
  File "C:\Users\Skoll\PycharmProjects\YouTube Downloader\venv\lib\site-packages\pytube\__main__.py", line 346, in title
    raise exceptions.PytubeError(
pytube.exceptions.PytubeError: Exception while accessing title of https://youtube.com/watch?v=AVYGCjwjUXg. Please file a bug report at https://github.com/pytube/pytube

And if I comment this : print(f"Titre: {youtube_video.title}") print(f"Nombre de vues : {youtube_video.views}"), I have this error : KeyError: 'streamingData'

I have version 12.1.3 of pytube. Can you help me please ?

g-sergey:
Seems YouTube never more provide data if client ==  'ANDROID'

[...]

Issue #1556, opened 2023-04-27
[BUG]KeyError: 'videoDetails' : pytube.exceptions.PytubeError: Exception while accessing title of https://youtube.com/watch?v=8C4Lo0a3YKY. Please file a bug report at https://github.com/pytube/pytube


q4udebobrota:
Traceback (most recent call last):
File "D:\pythonProjects\youtube-downloader\youtube-downloader-converter-master\code.py", line 35, in
stream = str(yt.streams.filter(progressive=True))
File "D:\pythonProjects\youtube-downloader\env\lib\site-packages\pytube_main_.py", line 296, in streams
return StreamQuery(self.fmt_streams)
File "D:\pythonProjects\youtube-downloader\env\lib\site-packages\pytube_main_.py", line 176, in fmt_streams
stream_manifest = extract.apply_descrambler(self.streaming_data)
File "D:\pythonProjects\youtube-downloader\env\lib\site-packages\pytube_main_.py", line 161, in streaming_data
return self.vid_info['streamingData']
KeyError: 'streamingData'

please help me with a solution

D1xx1:
> Traceback (most recent call last):
> File "D:\pythonProjects\youtube-downloader\youtube-downloader-converter-master\code.py", line 35, in
> stream = str(yt.streams.filter(progressive=True))
> File "D:\pythonProjects\youtube-downloader\env\lib\site-packages\pytube_main_.py", line 296, in streams
> return StreamQuery(self.fmt_streams)
> File "D:\pythonProjects\youtube-downloader\env\lib\site-packages\pytube_main_.py", line 176, in fmt_streams
> stream_manifest = extract.apply_descrambler(self.streaming_data)
> File "D:\pythonProjects\youtube-downloader\env\lib\site-packages\pytube_main_.py", line 161, in streaming_data
> return self.vid_info['streamingData']
> KeyError: 'streamingData'
> 
> please help me with a solution

https://github.com/pytube/pytube/issues/1553#issuecomment-1524956425
I try that solution but download speed leaves much to be desired.

ghost:
C:\Users\ss_ra\Downloads>python Youtube_download.py
Enter the link: https://www.youtube.com/watch?v=WXsD0ZgxjRw
The input link is https://www.youtube.com/watch?v=WXsD0ZgxjRw
Traceback (most recent call last):
[...]

Issue #1555, opened 2023-04-26
[BUG]  the module is showing error while extracting the title sometime 


NormVg:
**Before creating an issue**

Please confirm that you are on the latest version of pytube by installing from the source.
You can do this by running `python -m pip install git+https://github.com/pytube/pytube`.
Sometimes, the pypi library repository is not up to date, and your issue may have been fixed already!

**Describe the bug**

**To Reproduce**
Please provide the following information:
- The video or playlist url that is causing the error.
- The code where the problem is occurring.

**Expected behavior**

**Output**
If pytube raises an exception, please provide the full traceback for the exception.

**System information**
Please provide the following information:
- Python version (run `python --version`)
- Pytube version (run `print(pytube.__version__)` in python)
- Command used to install pytube

NormVg:
Traceback (most recent call last):
  File "/home/runner/Dusico/venv/lib/python3.10/site-packages/pytube/__main__.py", line 341, in title
    self._title = self.vid_info['videoDetails']['title']
KeyError: 'videoDetails'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/runner/Dusico/venv/lib/python3.10/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/runner/Dusico/venv/lib/python3.10/site-packages/flask/app.py", line 1822, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/runner/Dusico/venv/lib/python3.10/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/runner/Dusico/venv/lib/python3.10/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "main.py", line 43, in player
    data = tubedown.tubedata(id)
  File "/home/runner/Dusico/tubedown.py", line 18, in tubedata
    title = yt.title
  File "/home/runner/Dusico/venv/lib/python3.10/site-packages/pytube/__main__.py", line 346, in title
    raise exceptions.PytubeError(
pytube.exceptions.PytubeError: Exception while accessing title of https://youtube.com/watch?v=C3BE3jNeHtU. Please file a bug report at https://github.com/pytube/pytube

Konnected03:
I am also having issues with pytube 

Exception in Tkinter callback
Traceback (most recent call last):
[...]

PR #1559, opened 2023-04-27
Update innertube.py: changed ANDROID to WEB to fix KeyError

g-sergey:
This fixed "KeyError: 'videoDetails' ; Exception while accessing title of ...." (and similar) bug for me.  Seems it never more works for 'ANDROID'.

This is  a quck fix to make it work - not sure if it won't break anything in different place.

### DIFF
--- a/pytube/innertube.py
+++ b/pytube/innertube.py
@@ -75,7 +75,7 @@ _token_file = os.path.join(_cache_dir, 'tokens.json')
 
 class InnerTube:
     """Object for interacting with the innertube API."""
-    def __init__(self, client='ANDROID', use_oauth=False, allow_cache=True):
+    def __init__(self, client='WEB', use_oauth=False, allow_cache=True):
         """Initialize an InnerTube object.
 
         :param str client:


g-sergey:
Related bug reports:
#1558 #1557 #1556 #1555

Fire-Input:
Can confirm this fixed it for me.
I had a module that was heavily dependent on PyTube.
Thanks a lot!

g-sergey:
@Fire-Input I used another solution though: 
yt = YouTube(yt_link, use_oauth=True)

Added use_oauth param when creating YouTube object, and authorised it into my YouTube channel, so I have no to make any changes into the lib code before it's officially fixed. This creates __cache__ folder inside of the lib core directory, and puts tokens.json file inside.

kventinel:
Fix with WEB instead of ANDROID work's for me, but break video downloading, that's created infinite wait in this place. Second fix with oauth works well for me. Thanks:)

ArcticSlayy:
> Fix with WEB instead of ANDROID work's for me, but break video downloading, that's created infinite wait in this place. Second fix with oauth works well for me. Thanks:)

Same here.

AlejandroCardozoCode:
Where can I find the documentation to be able to set the credentials.json of use_oauth

Fire-Input:
@g-sergey Got it, Thank you.

g-sergey:
> Where can I find the documentation to be able to set the credentials.json of use_oauth

You don't need to, - it is created automatically. When you run script for the first time you will see prompt with a link and confirmation code.

AlanNaoto:
> @Fire-Input I used another solution though: yt = YouTube(yt_link, use_oauth=True)
> 
> Added use_oauth param when creating YouTube object, and authorised it into my YouTube channel, so I have no to make any changes into the lib code before it's officially fixed. This creates **cache** folder inside of the lib core directory, and puts tokens.json file inside.

Thanks, this is the better fix/workaround IMO compared to all those other PRs which change ANDROID to WEB. The latter, at least for me, just introduces a new error...

pranayp35:
changing **ANDROID to WEB** to fixes my  KeyError bug...

glubsy (changes requested):
This is not a proper fix. This is workaround that does not warrant this kind of patch.

On top of that, using the "WEB" client is known to cause heavy throttling.  And even then, it won't work anyway since the proper signatures are not computed. Have a look at what yt-dlp does if you actually want to patch the issue.

PR #1562, opened 2023-04-27
changed ANDROID to WEB to fix KeyError

isaacramsay:
Temp fix for a KeyError when calling Youtube().title and others that use vid_info

### DIFF
--- a/pytube/innertube.py
+++ b/pytube/innertube.py
@@ -75,7 +75,7 @@ _token_file = os.path.join(_cache_dir, 'tokens.json')
 
 class InnerTube:
     """Object for interacting with the innertube API."""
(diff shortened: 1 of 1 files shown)


ArcticSlayy:
When changing to web my program stops responding and the video isn't downloaded even when having the lowest resolution.

felipeucelli:
It would be interesting to switch to the WEB version, however most of the videos are encrypted, and it will be necessary to update cipher.py to correctly calculate "n", the current function does not support the new youtube JS.

player2I0:
after downloading one video, pytube throws that error: HTTP Error 403: Forbidden, and i can't download videos anymore.

DandyDrop:
The video is not downloading, even though the file is created (0 bytes file with video name)

PR #1619, opened 2023-05-11
Multiple updates (11 May 2023)

alexey-pelykh:


RONNCC:
Sorry what are the updates here ...?

alexey-pelykh:
@RONNCC quite reasonable question I may say :) I'll eventually try split it into at least several commits yet thus far the PR addresses:

- `Pipfile.lock` updates to actually work on Apple M1 with MacOS
- Improvement to `YouTube.vid_info()` since all the data is already there in the HTML, there's no need to call `InnerTube`
- Improvement to `Channel`: the data structure has changed
- Refactor `request.stream()`: it was a mess really with the Content-Range and Content-Length
- Various fixes: multiple tests were failing and some still are - seems like no-one have tried to run them for ages

piltom:
Hi, I wanted to mention that while this fixes both the issues mentioned in https://github.com/pytube/pytube/pull/1621 and https://github.com/pytube/pytube/issues/1620, it makes the downloads **significantly** slower. I timed it, it took 172 seconds for downloading an 8Mb video (versus 11.4 seconds for higher res version of the same video with the older code)

pbxforce:
As @piltom  said, this PR fixes both #1620 and #1621 issues with AgeRestrictionError and non-embedding video exception (which also gives AgeRestrictionError). But the speed is too much slower. It actually does not download video beyond 9MB. Downloading 720p video took 6 seconds with original code and file size was 26MB. Using this fix, downloading same 720p video took 2 minute and it downloaded only 9MB. It cut off process after 9MB.

alexey-pelykh:
@pbxforce thanks for the report, could you link the video so I can test? This was obviously not intended yet previous code wasn't working as intended by tests either - I think the chunk size is not sufficient and I'll look into that first thing tomorrow

alexey-pelykh:
Right, I see the issue with 9mb cutoff  - indeed my mistake. Yet I have no explanation why it is slower, most time is consumed by `self._sslobj.read(len, buffer)` as if YouTube itself is throttling the download.

alexey-pelykh:
Hm, I'm coming to conclusion that indeed that's the case - it's actually being throttled

piltom:
> Hm, I'm coming to conclusion that indeed that's the case - it's actually being throttled

Seems to be the same effect as using Innertube with WEB, which is the reason why in the original code it is avoided and the ANDROID player is preferred

alexey-pelykh:
> ANDROID player is preferred

It's `ANDROID_MUSIC` it seems from the code.

~Either way, `&alr=yes&cpn=-qPPiEfl3CfsJYd5&cver=2.20230512.01.00&range=0-9437184&rn=0&rbuf=0` contains some magic that avoids throttling.~

alexey-pelykh:
Oh well, it's better to remove reading from watch-html by default yet I've kept the `load_vid_info_from_watch_html()` method to allow #1620 and #1621 to be downloaded even if that's slower

alexey-pelykh:
I've measured that `ANDROID_MUSIC` client gives ~646kb/s and the via `ytInitialPlayerResponse` it's like ~50kb/s while via browser if to repeat streaming request via curl it's ~200kb/s

piltom:
> Oh well, it's better to remove reading from watch-html by default yet I've kept the `load_vid_info_from_watch_html()` method to allow #1620 and #1621 to be downloaded even if that's slower

In https://github.com/pytube/pytube/pull/1621 you can see my latest changes will still allow downloading the problematic videos faster than with watch-html too. There's a map from failure reason to player to use that should make it nicer to update in case of newer reasons or player changes.

The "update_vid_info_with" function in my PR does basically the same as the newly added "refresh_video_info" in yours. The argument to this function is then picked from the reasons map

alexey-pelykh:
@piltom cool, then I'll drop the not needed changes. Do you think `load_vid_info_from_watch_html()` should be kept or should it go?

pbxforce:
I think your last commit resolve the throttling issue. No speed throttle or size capping this time. It seems ANDROID and ANDROID_MUSIC both works fine. So far, i got no error. nice work

alexey-pelykh:
I'm splitting this PR into multiple:

- [x] Add `.editorconfig` via #1628
- [x] `Pipfile.lock` updates to actually work on Apple M1 with MacOS via #1629
- [ ] ~Improvement to `YouTube.vid_info()` since all the data is already there in the HTML, there's no need to call `InnerTube`~ yielding to #1621 by @piltom 
- [x] Improvement to `Channel`: the data structure has changed via #1630
- [x] Refactor `request.stream()`: it was a mess really with the `Content-Range` and `Content-Length` via #1632
- [x] Various fixes: multiple tests were failing and some still are - seems like no-one have tried to run them for ages via #1631

Issue #1620, opened 2023-05-11
[BUG] AgeRestrictedError with OAuth


yemregundogmus:
**Describe the bug**
When i try to download a youtube video it throws me a AgeRestrictionError. 

**To Reproduce**
Youtube URL: https://www.youtube.com/watch?v=JNW25pv6qRM
My Code: 
```python3
from pytube import YouTube
url = "https://www.youtube.com/watch?v=JNW25pv6qRM"
YouTube(url, use_oauth=True, allow_oauth_cache=False).streams.filter(only_audio=True).first().download(filename='test_vid.mp4')
```
- The code where the problem is occurring.
```
─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ C:\Users\SD\AppData\Local\Temp\ipykernel_20012\2311628616.py:9 in <cell line: 2>                 │
│                                                                                                  │
│ [Errno 2] No such file or directory:                                                             │
│ 'C:\\Users\\SD\\AppData\\Local\\Temp\\ipykernel_20012\\2311628616.py'                            │
│                                                                                                  │
│ C:\Users\SD\AppData\Local\Temp\ipykernel_20012\3322474857.py:160 in pipeline                     │
│                                                                                                  │
│ [Errno 2] No such file or directory:                                                             │
│ 'C:\\Users\\SD\\AppData\\Local\\Temp\\ipykernel_20012\\3322474857.py'                            │
│                                                                                                  │
│ C:\Users\SD\AppData\Local\Temp\ipykernel_20012\3322474857.py:52 in download_youtube_video        │
│                                                                                                  │
│ [Errno 2] No such file or directory:                                                             │
│ 'C:\\Users\\SD\\AppData\\Local\\Temp\\ipykernel_20012\\3322474857.py'                            │
│                                                                                                  │
│ C:\ProgramData\Anaconda3\lib\site-packages\pytube\__main__.py:296 in streams                     │
│                                                                                                  │
│   293 │   │   :rtype: :class:`StreamQuery <StreamQuery>`.                                        │
│   294 │   │   """                                                                                │
│   295 │   │   self.check_availability()                                                          │
│ ❱ 296 │   │   return StreamQuery(self.fmt_streams)                                               │
│   297 │                                                                                          │
│   298 │   @property                                                                              │
│   299 │   def thumbnail_url(self) -> str:                                                        │
│                                                                                                  │
[...]

PR #1621, opened 2023-05-11
Embed playback issue: When embed playback is blocked, pytube would detect it as age restriction

piltom:
Hello. This morning while running some test downloads I ran into an issue with the following video: https://www.youtube.com/watch?v=e8RcQoGY4OE

Pytube would detect it as Age restricted even though it is not. After some digging I found it was due to the video having embed playback blocked.  

The current solution in this PR will check the if reason for no playability is present in the newly introduced map and obtain which player to use to work around it.

This solution could be problematic in the future for two reasons:
 - Are the reasons for no playability always in english?
 - Is it often that they change the text? If so, maybe we would need to interpret the reason with searching some string in it.. (like "embed" or "website")

Here are some links for you to test:
1. The link that is not allowed in embedded: https://www.youtube.com/watch?v=e8RcQoGY4OE  which should now be downloaded correctly (previously would throw age restriction exception)
2. An Eminem song that is age restricted and will properly throw the exception: https://www.youtube.com/watch?v=mQvteoFiMlg
3. Pick any other video from youtube and it should still work as originally

### DIFF
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -31,7 +31,8 @@ class YouTube:
         on_complete_callback: Optional[Callable[[Any, Optional[str]], None]] = None,
         proxies: Dict[str, str] = None,
         use_oauth: bool = False,
-        allow_oauth_cache: bool = True
+        allow_oauth_cache: bool = True,
+        innertube_client: str = 'ANDROID'
     ):
         """Construct a :class:`YouTube <YouTube>`.
 
@@ -51,6 +52,11 @@ class YouTube:
         :param bool allow_oauth_cache:
             (Optional) Cache OAuth tokens locally on the machine. Defaults to True.
             These tokens are only generated if use_oauth is set to True as well.
+        :param str innertube_client:
+            (Optional) Which client to use for interacting with youtube. String has to be one of:
+            'WEB', 'ANDROID', 'IOS', 'WEB_EMBED', 'ANDROID_EMBED', 'IOS_EMBED', 'WEB_MUSIC',
+            'ANDROID_MUSIC', 'IOS_MUSIC', 'WEB_CREATOR', 'ANDROID_CREATOR', 'IOS_CREATOR',
+            'MWEB', 'TV_EMBED'
         """
         self._js: Optional[str] = None  # js fetched by js_url
         self._js_url: Optional[str] = None  # the url to the js, parsed from watch html
@@ -87,6 +93,8 @@ class YouTube:
 
         self.use_oauth = use_oauth
         self.allow_oauth_cache = allow_oauth_cache
+
+        self.innertube_client = innertube_client
 
     def __repr__(self):
         return f'<pytube.__main__.YouTube object: videoId={self.video_id}>'
@@ -154,11 +162,8 @@ class YouTube:
     @property
     def streaming_data(self):
         """Return streamingData from video info."""
-        if 'streamingData' in self.vid_info:
-            return self.vid_info['streamingData']
-        else:
-            self.bypass_age_gate()
-            return self.vid_info['streamingData']
+        return self.vid_info['streamingData']
+
 
     @property
     def fmt_streams(self):
@@ -232,6 +237,32 @@ class YouTube:
             elif status == 'LIVE_STREAM':
                 raise exceptions.LiveStreamError(video_id=self.video_id)
 
+    def check_innertube_availability(self, innertube_response):
+        """Check whether video is available by analysing the innertube response
+
+        Raises different exceptions based on why the video is unavailable,
+        otherwise does nothing.
+        """
+        status = innertube_response['playabilityStatus']['status']
+        if status == 'OK':
+            return
+
+        reason = innertube_response['playabilityStatus']['reason']
+
+        if reason == 'Sorry, this content is age-restricted' or reason == 'This video may be inappropriate for some users.' or reason == 'Sign in to confirm your age':
+            raise exceptions.AgeRestrictedError(video_id=self.video_id)
+
+        if reason == 'Playback on other websites has been disabled by the video owner':
+            raise exceptions.EmbedPlaybackDisabled(video_id=self.video_id)
+
+        # if we got this far and none of the reasons apply and login is required
+        # then we raise a video private exception
+        if status == 'LOGIN_REQUIRED':
+            raise exceptions.VideoPrivate(video_id=self.video_id)
+
+        # if the status was not login required, then we raise a more generic video unavailable exception
+        raise exceptions.VideoUnavailable(video_id=self.video_id)
+
     @property
     def vid_info(self):
         """Parse the raw vid info and return the parsed result.
@@ -241,29 +272,13 @@ class YouTube:
         if self._vid_info:
             return self._vid_info
 
-        innertube = InnerTube(use_oauth=self.use_oauth, allow_cache=self.allow_oauth_cache)
+        innertube = InnerTube(client=self.innertube_client, use_oauth=self.use_oauth, allow_cache=self.allow_oauth_cache)
 
         innertube_response = innertube.player(self.video_id)
+        self.check_innertube_availability(innertube_response)
+
         self._vid_info = innertube_response
         return self._vid_info
-
-    def bypass_age_gate(self):
-        """Attempt to update the vid_info by bypassing the age gate."""
-        innertube = InnerTube(
-            client='ANDROID_EMBED',
-            use_oauth=self.use_oauth,
-            allow_cache=self.allow_oauth_cache
-        )
-        innertube_response = innertube.player(self.video_id)
-
-        playability_status = innertube_response['playabilityStatus'].get('status', None)
-
-        # If we still can't access the video, raise an exception
-        # (tier 3 age restriction)
-        if playability_status == 'UNPLAYABLE':
-            raise exceptions.AgeRestrictedError(self.video_id)
-
-        self._vid_info = innertube_response
 
     @property
     def caption_tracks(self) -> List[pytube.Caption]:
--- a/pytube/exceptions.py
+++ b/pytube/exceptions.py
@@ -52,6 +52,19 @@ class VideoUnavailable(PytubeError):
     def error_string(self):
         return f'{self.video_id} is unavailable'
 
+class EmbedPlaybackDisabled(VideoUnavailable):
+    """Playback on other websites has been disabled by the video owner."""
+    def __init__(self, video_id: str):
+        """
+        :param str video_id:
+            A YouTube video identifier.
+        """
+        self.video_id = video_id
+        super().__init__(self.video_id)
+
+    @property
+    def error_string(self):
+        return f"{self.video_id} can only be reproduced on the youtube website."
 
 class AgeRestrictedError(VideoUnavailable):
     """Video is age restricted, and cannot be accessed without OAuth."""


piltom:
Here is some minimal code to reproduce the issue before the patch:
```python
from pytube import YouTube as PyYouTube

# Video that is not allowed as embedded
yt = PyYouTube("https://www.youtube.com/watch?v=e8RcQoGY4OE")
yt.check_availability()
stream = yt.streams.first()
filename = stream.download("./")

# normal video
yt = PyYouTube("https://www.youtube.com/watch?v=wMRbSKjEtrc")
yt.check_availability()
stream = yt.streams.first()
filename = stream.download("./")

# age restricted video
yt = PyYouTube("https://www.youtube.com/watch?v=mQvteoFiMlg")
yt.check_availability()
stream = yt.streams.first()
filename = stream.download("./")
```

pbxforce:
I tried this commit. It worked for some videos but not for all. I still got error of age restricted when they are not restricted. Check this url: "https://www.youtube.com/shorts/DYVUwiB6fLM"
and this one: "https://www.youtube.com/watch?v=QBvzEcXSUjg"

both URLs are not age restricted and no embedding blocked but still giving exception of AgeRestrictedError

piltom:
> I tried this commit. It worked for some videos but not for all. I still got error of age restricted when they are not restricted. Check this url: "https://www.youtube.com/shorts/DYVUwiB6fLM" and this one: "https://www.youtube.com/watch?v=QBvzEcXSUjg"
> 
> both URLs are not age restricted and no embedding blocked but still giving exception of AgeRestrictedError

Hey, there is also this PR https://github.com/pytube/pytube/pull/1619 that seems to change the way data is obtained at the start, I haven't tried it but you can give it a shot.

 My solution was for this particular case of embed blocking, but it seems there's more cases in which this happens so it should be generalized (I don't like the way I structured this solution particularly).

pbxforce:
> > I tried this commit. It worked for some videos but not for all. I still got error of age restricted when they are not restricted. Check this url: "https://www.youtube.com/shorts/DYVUwiB6fLM" and this one: "https://www.youtube.com/watch?v=QBvzEcXSUjg"
> > both URLs are not age restricted and no embedding blocked but still giving exception of AgeRestrictedError
> 
> Hey, there is also this PR #1619 that seems to change the way data is obtained at the start, I haven't tried it but you can give it a shot.
> 
> My solution was for this particular case of embed blocking, but it seems there's more cases in which this happens so it should be generalized (I don't like the way I structured this solution particularly).

I'll give #1619 a try. BTW your commit does worked for embedding blocked videos. I tried on some more videos that have embedding blocked, this commit can bypass that embedding block gate.

piltom:
@pbxforce I have updated the code and now both your links work. It is also a lot faster that the other PR mentioned previously.

What I added now is a mapping from unplayability reason to player to use to bypass said reason. This way, if you encounter new problems you can update this map (check the reasons.py map). You can extract the reason by adding a print statement in the newly updated code for vid_info

piltom:
The mapping for now looks like this:

```python
PLAYER_FOR_REASON = {
  'Playback on other websites has been disabled by the video owner': 'ANDROID',
  'This video is not available': 'ANDROID',
  'This video is unavailable': 'ANDROID',
}
```
Which makes it seem redundant because all the vales are ANDROID. Could come in handy if there are other reasons that are not bypassed by ANDROID, but I don't have much experience with youtube to tell. 

This could also eventually be updated to have a list of players for each key, so that the code can try all of them until one succeeds.

pbxforce:
This is good. I somewhat fixed it by changing 'client=ANDROID_EMBED  to 'client=ANDROID' but it was not totally fixed. Your way is much better and getting playabilityStatus reason is also good to know for future errors. Thank you

pbxforce:
This PR resolve issues related to #1620 of AgeRestrictionError for non-restricted videos. So far, everything looks good.

pbxforce (approved):
This PR fixes the issue for me related to #1620 of AgeRestrictionError for non-restricted and non-embedding videos (which also gives exception of AgeRestrictionError).

piltom:
@nficano / @glubsy/ @tfdahlin / @RONNCC  could I get a review/feedback on this?

The age restriction reason should probably be moved to the map too no?

If android will fix all reasons then probably we don't even need a map, also it would be nice to include the reason for unplayability in the exception, wouldn't you agree?

felipeucelli:
I tested the code and it manages to get the video when it is prevented from embedding. But for some reason the **ANDROID** client is limiting the validity of the url to about 30 seconds, after that time, a 403 error is generated.

We could use the **IOS** client, which has no throttling. But it returns few stream options.

I think the best thing would be if we could make the **ANDROID** client work normally to use it as the default in **innertube.py**.

pytube/__main__.py
```diff
@@ -247,6 +254,21 @@ def vid_info(self):
...
+            allow_cache=self.allow_oauth_cache
+        )
+
+        innertube_response = innertube.player(self.video_id)
+        playability_status = innertube_response['playabilityStatus'].get('status', None)
+        # If we still can't access the video, raise an exception
+        if playability_status == 'UNPLAYABLE':
+            raise exceptions.VideoUnavailable(self.video_id)
+        self._vid_info = innertube_response
+
```
glubsy: At first glance I would say this entire method should be part of `YouTube.vid_info`, no need for a separate detour like that? `vid_info` could take a default `client` parameter?

pytube/reasons.py
```diff
@@ -0,0 +1,7 @@
+# This dictionary can be used for mapping playability issues' reasons
+# to players to use to bypass said reason.
+PLAYER_FOR_REASON = {
+  'Playback on other websites has been disabled by the video owner': 'ANDROID',
+  'This video is not available': 'ANDROID',
+  'This video is unavailable': 'ANDROID',
+}
```
glubsy: This module's name and this variable name are not very clear. I don't think it needs its own separate module.
This could be better documented with nested key/value pairs, ie 
```python
FALLBACK_CLIENTS = [
    {
        'error_message': 'Playback on other websites has been disabled by the video owner',
        'client': 'ANDROID'
    },
    ...
]
```

glubsy (commented):
Not sure what to think of this workaround. It might work, but forcing whatever client just because the first one did not work sounds like a poor approach.
If anything the "client" application should handle that, so perhaps the only thing worth doing here would be to throw the appropriate exception instead of trying to work around whatever data you get in the "library" part of the code.

pytube/reasons.py
glubsy: Then again I don't think this should be handled here.

RONNCC:
aligned /w glubsy ... slightly scared if this would break other things 🤔

RONNCC:
would you have any insight into a generic solution across clients perhaps @piltom ?

piltom:
> Not sure what to think of this workaround. It might work, but forcing whatever client just because the first one did not work sounds like a poor approach. If anything the "client" application should handle that, so perhaps the only thing worth doing here would be to throw the appropriate exception instead of trying to work around whatever data you get in the "library" part of the code.

Alright, this approach was just a generalization of the original `bypass_age_gate` method that was originally assuming that the problem was age restriction no matter what, then trying ANDROID_EMBED and finally throwing an age restriction excpetion (incredibly confusing when it is not the case).

Would you agree then to remove the `bypass_age_gate` method and instead throw an appropriate exception that contains the reason returned by Youtube? This exception can also then contain a suggestion to use the new client parameter to vid_info that @glubsy suggested.




> I tested the code and it manages to get the video when it is prevented from embedding. But for some reason the **ANDROID** client is limiting the validity of the url to about 30 seconds, after that time, a 403 error is generated.

I will try to reproduce that and see if there's something I can do

piltom:
As a matter of fact, it seems that `bypass_age_gate` is being called always in the original code, so the Innertube client being used by default is technically ANDROID_EMBED

glubsy:
> Would you agree then to remove the `bypass_age_gate` method and instead throw an appropriate exception that contains the reason returned by Youtube? This exception can also then contain a suggestion to use the new client parameter to vid_info that @glubsy suggested.
>

That sounds much more reasonable to me, although the exception should not really care about workarounds, only the calling code should then suggest what to do in that case.

> I tested the code and it manages to get the video when it is prevented from embedding. But for some reason the **ANDROID** client is limiting the validity of the url to about 30 seconds, after that time, a 403 error is generated.

As far as I know, this has been going on since early april of 2023. This PR does not aim at solving this issue.

piltom:
I'll push a cleaner version later today. In the meantime here is some playability status details for each client, see how many of them differ in both status and reason depending on the client. Also it seems that the original age bypass does not work anymore:

Format is client: (status, reason)

```
Embedded blocked (https://www.youtube.com/watch?v=e8RcQoGY4OE)
{
 'WEB': ('OK', ''),
 'ANDROID': ('OK', ''),
 'IOS': ('OK', ''),
 'WEB_EMBED': ('UNPLAYABLE', 'Video unavailable'),
 'ANDROID_EMBED': ('UNPLAYABLE', 'Playback on other websites has been disabled by the video owner'),
 'IOS_EMBED': ('OK', ''),
 'WEB_MUSIC': ('UNPLAYABLE', 'This video is not available'),
 'ANDROID_MUSIC': ('UNPLAYABLE', 'This video is not available'),
 'IOS_MUSIC': ('UNPLAYABLE', 'This video is not available'),
 'WEB_CREATOR': ('OK', ''),
 'ANDROID_CREATOR': ('OK', ''),
 'IOS_CREATOR': ('OK', ''),
 'MWEB': ('OK', ''),
 'TV_EMBED': ('UNPLAYABLE', 'Playback on other websites has been disabled by the video owner')
}

Normal video (https://www.youtube.com/watch?v=wMRbSKjEtrc)
{
 'WEB': ('OK', ''),
 'ANDROID': ('OK', ''),
 'IOS': ('OK', ''),
 'WEB_EMBED': ('OK', ''),
 'ANDROID_EMBED': ('OK', ''),
 'IOS_EMBED': ('OK', ''),
 'WEB_MUSIC': ('UNPLAYABLE', 'This video is not available'),
 'ANDROID_MUSIC': ('UNPLAYABLE', 'This video is not available'),
 'IOS_MUSIC': ('UNPLAYABLE', 'This video is not available'),
 'WEB_CREATOR': ('OK', ''),
 'ANDROID_CREATOR': ('OK', ''),
 'IOS_CREATOR': ('OK', ''),
 'MWEB': ('OK', ''),
 'TV_EMBED': ('OK', '')
}

Age restricted (https://www.youtube.com/watch?v=mQvteoFiMlg)
{
 'WEB': ('LOGIN_REQUIRED', 'Sign in to confirm your age'),
 'ANDROID': ('LOGIN_REQUIRED', 'This video may be inappropriate for some users.'),
 'IOS': ('LOGIN_REQUIRED', 'This video may be inappropriate for some users.'),
 'WEB_EMBED': ('UNPLAYABLE', 'Sorry, this content is age-restricted'),
 'ANDROID_EMBED': ('UNPLAYABLE', 'Sorry, this content is age-restricted'),
 'IOS_EMBED': ('UNPLAYABLE', 'Sorry, this content is age-restricted'),
 'WEB_MUSIC': ('LOGIN_REQUIRED', 'Sign in to confirm your age'),
 'ANDROID_MUSIC': ('LOGIN_REQUIRED', 'Sign in to confirm your age'),
 'IOS_MUSIC': ('LOGIN_REQUIRED', 'Sign in to confirm your age'),
 'WEB_CREATOR': ('LOGIN_REQUIRED', 'Sign in to confirm your age'),
 'ANDROID_CREATOR': ('LOGIN_REQUIRED', 'Sign in to confirm your age'),
 'IOS_CREATOR': ('LOGIN_REQUIRED', 'Sign in to confirm your age'),
 'MWEB': ('LOGIN_REQUIRED', 'Sign in to confirm your age'),
 'TV_EMBED': ('UNPLAYABLE', 'This video is unavailable')
 }
```

piltom:
@glubsy @RONNCC  what do you think of the current solution? I force pushed a squashed commit :)

The pytube constructor takes the optional innertube client argument (vid_info is a property getter, it cannot take arguments). Then the vid_info getter will check for availability on the innertube response. There's another check for availability in the code but it seems to only be used on the watch_html... should this maybe be updated or generalized?

Doing the check for availability in the vid_info getter might be too soon? Where would you place that instead?

I can also add a new feature for troubleshooting that will analyze the problems on each client to suggest the user what the problem is. If you check my previous comment in the PR, in some cases the youtube reasons are vague and overlap with other problems. This would be something like:

```python
def troubleshoot_url(self):
    # go through all clients to find a good reason why the link does not work
    # maybe raise the correct exception?
```

PR #1623, opened 2023-05-12
updted extract channel_name pattern

deduble:
channel_name pattern doesn't even consider youtube URLs starting with `www`. New pattern also includes those as well as new `youtube.com/*` channel names.

### DIFF
--- a/pytube/extract.py
+++ b/pytube/extract.py
@@ -167,20 +167,12 @@ def channel_name(url: str) -> str:
     :returns:
         YouTube channel name.
     """
-    patterns = [
-        r"(?:\/(c)\/([%\d\w_\-]+)(\/.*)?)",
-        r"(?:\/(channel)\/([%\w\d_\-]+)(\/.*)?)",
-        r"(?:\/(u)\/([%\d\w_\-]+)(\/.*)?)",
-        r"(?:\/(user)\/([%\w\d_\-]+)(\/.*)?)"
-    ]
-    for pattern in patterns:
-        regex = re.compile(pattern)
-        function_match = regex.search(url)
-        if function_match:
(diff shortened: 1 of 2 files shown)


RONNCC:
could you add some tests to ensure we don't break the existing pattern matching?

deduble:
Isn't there already tests in [test_channel.py]( https://github.com/pytube/pytube/blob/master/tests/contrib/test_channel.py)
Would you like me to add more?

RONNCC:
from my understanding the previous urls did not consider www patterns -> hence we'd like to add new test cases to correspond to these

deduble:
Done Added some tests. Never mind it wasn't about `www`'s. New URIs could contain @ that's why it couldn't read.
However more important channel pattern to match was currently being used is `(https://www.)youtube.com/@channel_name` was not being matched. Passes for old and new test cases I added.

pytube/extract.py
```diff
@@ -167,20 +167,12 @@ def channel_name(url: str) -> str:
     :returns:
         YouTube channel name.
     """
-    patterns = [
-        r"(?:\/(c)\/([%\d\w_\-]+)(\/.*)?)",
-        r"(?:\/(channel)\/([%\w\d_\-]+)(\/.*)?)",
-        r"(?:\/(u)\/([%\d\w_\-]+)(\/.*)?)",
-        r"(?:\/(user)\/([%\w\d_\-]+)(\/.*)?)"
```
RONNCC, line 174: can we keep the structure of breaking out the patterns.

Makes it much easier to see what categories of URLs are supported (and is easier to debug) -- versus the combined regex below)

deduble:
I already immediately dropped this library while developing. most stuff is already broken esp. in contrib. sorry I am not further going to help on this.

Issue #1453, opened 2023-01-13
Taking too long when extracting streams from a video pytube 12.1.2


IamBrazilian:
**Yesterday pytube fetched streams for about 5 seconds now it's 18 seconds upwards. It's taking too long.
Can anyone tell me what the problem is?**

```
from pytube import YouTube
from time import time

yt = YouTube("https://www.youtube.com/watch?v=mUs97qXjw1M")

time0 = time()
print(yt.streams)
time1 = time() - time0

print(time1)
```

```
18.189778327941895
```

Python 3.9
Pytube 12.1.2

AtecAtca:
just solved this problem.

Inside the parser.py - def throttling_array_split():
there is a loop that runs for about 180,000 iterations and increased more than 10 seconds

Decision:

open pytube/_ _main_ _.py
go to line: def fmt_streams(self):

comment or delete this code:

# If the cached js doesn't work, try fetching a new js file
        # https://github.com/pytube/pytube/issues/1054
        #try:
        #
        #    extract.apply_signature(stream_manifest, self.vid_info, self.js)
        #
        #except exceptions.ExtractError:
        #    # To force an update to the js file, we clear the cache and retry
        #    self._js = None
        #    self._js_url = None
        #    pytube.__js__ = None
        #    pytube.__js_url__ = None
        #    extract.apply_signature(stream_manifest, self.vid_info, self.js)


Python 3.11.1
Pytube 12.1.2

IamBrazilian:
This really worked, the search is almost instantaneous, thank you very much!

huynguyencong:
This works. Thank you!

frenchfaso:
Awesome!
Thanks :+1:

Mchoga:
i just tried it and it seems not to be working for me. am i missing out something or youtube just messed up again

PR #1638, opened 2023-05-18
#1453 Workaround

AKozhuhov:


glubsy:
Please keep the reference to the Github issue in the PR comment, so that we can find the original issue easily.

I'm not convinced this is the right fix for the issue (the issue is not clearly defined either). You're skipping applying the signature cipher entirely, which has a negative effect down the line.

By "negative effect" I mean that the resulting URIs for streams will not work at all, or at best be throttled heavily.

Issue #1270, opened 2022-03-31
Unexpected renderer encountered. 'reelShelfRenderer'


bakarbyikov:
**Warning shows sometimes when searching some videos**

```
Unexpected renderer encountered.
Renderer name: dict_keys(['reelShelfRenderer'])
Search term: The Lumineers - Ho Hey
```

**code:**
```
from pytube import Search

song = "The Lumineers - Ho Hey"
for i in range(100):
    video = Search(song).results[0]
    print(video.title)
```

**output:**
```
The Lumineers - Ho Hey (Official Video)
The Lumineers - Ho Hey (Official Video)
The Lumineers - Ho Hey (Official Video)
Unexpected renderer encountered.
Renderer name: dict_keys(['reelShelfRenderer'])
Search term: The Lumineers - Ho Hey
Please open an issue at https://github.com/pytube/pytube/issues and provide this log output.
The Lumineers - Ho Hey (Official Video)
Unexpected renderer encountered.
Renderer name: dict_keys(['reelShelfRenderer'])
Search term: The Lumineers - Ho Hey
Please open an issue at https://github.com/pytube/pytube/issues and provide this log output.
The Lumineers - Ho Hey (Official Video)
The Lumineers - Ho Hey (Official Video)
The Lumineers - Ho Hey (Official Video)
The Lumineers - Ho Hey (Official Video)
...
```
**System information**

- Python version: Python 3.8.10
- Pytube version: 12.0.0
- Command used to install pytube: `python -m pip install git+https://github.com/pytube/pytube`

gmanolesi:
I have also ran into this issue. Will this be solved?

rgryta:
I've also encountered this problem.

After some digging I found that reelShelfRenderer is responsible for displaying YouTube Shorts. Since they're quite new to YT it seems like support for Shorts wasn't yet added to pytube. That being said, I tried to tackle this issue and fixed it on my fork for pytube. Added pull request  #1360 for review. You can of course wait until it's merged, or install it directly from github via pip.

ar852:
Until it is fixed, is it possible to suppress this output?

rgryta:
> Until it is fixed, is it possible to suppress this output?

Easiest way would be to install pytube with my fix I guess (comment above). Or you can take a look at how did it and fix it on your own.
Command to install pytube from my git repo is something like this:
`pip3 install git+https://github.com/rgryta/pytube.git@main`

[...]

PR #1642, opened 2023-05-20
Update search.py

colinpannikkat:
Prevent render bug for reelShelfRenderer and showingResultsforRenderers


ojh050118:
Does this PR fix #1661 and #1270?

colinpannikkat:
I think so, the reelShelfRenderer error happens when the program parses shorts videos which it can't handle. The code I added removes the warning from the program for reelShelfRenderer and showingResultsForRenderer. It was just a quick fix I made so I didn't do any in-depth testing, but within my usage it removed any warnings for both of those renderers.

PR #1649, opened 2023-05-24
Update cipher.py

mhi-iranmanesh:


deepaerial:
Is this PR fixes `pytube.exceptions.RegexMatchError: get_throttling_function_name: could not find match for multiple` exception?

Issue #1627, opened 2023-05-14
Unable to get videos on Channel [BUG]


mwendigg:
**Describe the bug**
I am unable to retrieve the URLs for videos on a channel. Other functionalities for channel objects continue to work.

**To Reproduce**
c = Channel('https://www.youtube.com/c/ProgrammingKnowledge/videos')
c.videos

c.channel_name

c.html

c = Channel('https://www.youtube.com/c/ProgrammingKnowledge')
c.videos

**Expected behavior**
It is supposed to return URLs of the videos on the channel.

**Output**
Constantly only [] for the .videos functionality (empty). Normal for others

**System information**
Please provide the following information:
- Python version Python 3.11.2
- Pytube version 15.0.0
- Command used to install pytube: python -m pip install git+https://github.com/pytube/pytube

akincan-kilic:
Same issue here.

BuildWithData:
@akincan-kilic @mwendigg  patch [this](https://github.com/pytube/pytube/pull/1409)

ttrreevvoorr:
This issue still exists.  The following code prints an empty array:
```python3
from pytube import Channel

channel_url = 'https://www.youtube.com/c/ProgrammingKnowledge/videos'
channel = Channel(channel_url)

print(channel.videos)
```
This example is from the docs: https://pytube.io/en/latest/user/channel.html

Perhaps something changed on YouTube's side?

BuildWithData:
> This issue still exists. The following code prints an empty array:
> 
> ```python
> from pytube import Channel
> 
> channel_url = 'https://www.youtube.com/c/ProgrammingKnowledge/videos'
> channel = Channel(channel_url)
> 
> print(channel.videos)
> ```
> 
> This example is from the docs: https://pytube.io/en/latest/user/channel.html
> 
> Perhaps something changed on YouTube's side?

have you patched your code ? I'm running version 15.0.0 and patched with this [pull](https://github.com/pytube/pytube/pull/1409) and it works, there are 3413 videos in total...

if you are having troubles patching your code, I made a [tutorial](https://youtu.be/TAL0G9IXdmw), just go to section at 02:57... hope this helps

nuhmanpk:
This issue still exists, 


```ImportError                               Traceback (most recent call last)
[<ipython-input-89-0a9798f3d5e9>](https://localhost:8080/#) in <cell line: 1>()
----> 1 from pytube import Channel
      2 
[...]

Issue #1546, opened 2023-04-20
[ BUG ] Cant get videos from channel


Godwhitelight:
It seems like something broke in the channel video scraping.
```py
channel = Channel('https://www.youtube.com/channel/UCvLswWFz2WGhdJ9cBftmhcA')
print(channel.channel_name, channel.channel_id)
for i in channel.videos:
    print(i)
```
this prints only the first print.
It does not find any videos. But the channel has videos 100%

TymoteuszChatys:
same error here

borgnetzwerk:
Same error, tried every possible combination of link, prefix, suffix, cid, name, ...

CyberGrim:
I'm assuming something changed on the YouTube side of things and we'll need to wait for an update to fix it.

Nothing has changed in my code, and I'm using the same version of PyTube that's always worked, but while I'm able to retrieve channel information, the class instance has nothing in the .video attribute.

Godwhitelight:
ive found their fix and i fixed it for myself ill send later

PR #1422, opened 2022-11-22
Fix empty channel videos, live and shorts

felipeucelli:
### Added support for shorts #1418 

Youtube changed html page, post data and json tree.

Now it is necessary to send a VisitorData in the continuation post.

Added `.shorts` property to get them from channels.

### Fixed empty video list #1408 
When youtube changed the shorts tab it also changed the json tree of the videos.

### Adding regex support to get new channel url style #1420 
In the recent youtube update it allowed channels to add custom names in the url 
"**@**" is now used before the channel name.

### Updated getting active tabs. [suggestion](https://github.com/pytube/pytube/pull/1422/files#r1030883623)

As youtube does not have a fixed index of each tab, 
we check if the url in the json tree of each tab is equal to the desired url.

### Added support for live. [solicitation](https://github.com/pytube/pytube/pull/1422/files#r1030118335)

Added `.live` property to get them from channels.

Issue #1410, opened 2022-11-02
Channel List Now Empty


ItsJoeTurner:
When using the Channel() method, it returns no videos. This is because of the new changes to YouTube over the past week. Has anyone found a fix, or can one by implemented ASAP as I'm hosting w/ PythonAnywhere so editing the package, or using a pull request, is tricky.

thedtvn:
same

thedtvn:
playlist to

parvez-cet21:
@ItsJoeTurner  Please refer to #1409 The PR is in Approval stage.

PR #1655, opened 2023-05-27
fix Channel._extract_videos (#1408, #1410, #1416, #1422, #1546, #1627)

heksbeks:
fixes also parsing for shorts, lives and topic channels (#1093)

This is a much simpler parser and will also likely be working when youtube changes json structure again since it does not use a full path.


TheLastochka:
Thank you

PR #1683, opened 2023-06-24
fix: RegexMatchError: get_throttling_function_name, RegexMatchError: channel_name, empty channel video

mdaniyalk:


Delafinn:
Can confirm this worked flawlessly. 
Thank you!

PR #1685, opened 2023-06-24
Update cipher.py

oncename:
FIX cipher.py

    function_patterns = [
        # https://github.com/ytdl-org/youtube-dl/issues/29326#issuecomment-865985377
        # https://github.com/yt-dlp/yt-dlp/commit/48416bc4a8f1d5ff07d5977659cb8ece7640dcd8
        # var Bpa = [iha];
        # ...
        # a.C && (b = a.get("n")) && (b = Bpa[0](b), a.set("n", b),
        # Bpa.length || iha("")) }};
        # In the above case, `iha` is the relevant function name
        r'a\.[a-zA-Z]\s*&&\s*\([a-z]\s*=\s*a\.get\("n"\)\)\s*&&.*?\|\|\s*([a-z]+)',
        r'\([a-z]\s*=\s*([a-zA-Z0-9$]+)(\[\d+\])?\([a-z]\)',
    ]

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,8 +74,8 @@ To download a video using the library in a script, you'll need to import the You
 
 ```python
  >>> from pytube import YouTube
- >>> YouTube('https://youtu.be/2lAe1cqCOXo').streams.first().download()
- >>> yt = YouTube('http://youtube.com/watch?v=2lAe1cqCOXo')
+ >>> YouTube('https://youtu.be/AXkbqdJFbcE').streams.first().download()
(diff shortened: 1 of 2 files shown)


glubsy (changes requested):
1. Why edit the Readme in such a way?
2. Why remove the docstring?
3. Trying to silently ignore the error will not fix the error
4. Returning an empty list is NOT what the get_transform_object method is expected to do
5. This is a duplicate PR
6. Your left unnecessary comments
7. No unit tests to corroborate the regex change

oncename:
https://github.com/oncename/pytube

Issue #1728, opened 2023-07-17
pytube.exceptions.RegexMatchError: get_transform_object: could not find match for var for={(.*?)};


HajoBrandt:
While downloading a video using the PyTube library using this code: yt.streams.get_highest_resolution().download("PATH", f"PATH.mp4") I get an error like this:

raise RegexMatchError(caller="get_transform_object", pattern=pattern)pytube.exceptions.RegexMatchError: get_transform_object: could not find match for var for={(.*?)};

I've seen a lot of fixes on stackoverflow and in the git repo of Pytube, but they seem to go into different parts of cypher.py. I would like to know how I could alternate get_transform_object class in cypher.py to match the regex check.

kar3399:
I'm getting the same error.

untilhamza:
I getting the same error too!

ejrtks1020:
I getting the same error

giulianoaccorsi:
I'm getting the same error too

ricardofriba:
same here

seamcarving:
I'm getting the same error

yiguid:
same here...

DavidOsparks:
Same here +1

ethanjoby:
same here

hp561:
Sammmeeee! I hope we can get this fixed 🙏🏽

luzhonghao1989:
same here

untilhamza:
## Here is quick fix in the meantime
https://stackoverflow.com/a/76718022/13889098

->  in file .venv/lib/python3.10/site-packages/pytube/cipher.py
-> find the method get_transform_object
```
def get_transform_object(js: str, var: str) -> List[str]:
    pattern = r"var %s={(.*?)};" % re.escape(var)
    logger.debug("getting transform object")
    regex = re.compile(pattern, flags=re.DOTALL)
    transform_match = regex.search(js)
    
    if not transform_match:
        # i commented out the line raising the error
        # raise RegexMatchError(caller="get_transform_object", pattern=pattern)
        return []  # Return an empty list if no match is found

    return transform_match.group(1).replace("\n", " ").split(", ")
```

awmie:
> ## Here is quick fix in the meantime
> https://stackoverflow.com/a/76718022/13889098
> 
> -> in file .venv/lib/python3.10/site-packages/pytube/cipher.py -> find the method get_transform_object
> 
> ```
> def get_transform_object(js: str, var: str) -> List[str]:
[...]

PR #1731, opened 2023-07-19
Fix for RegexMatchError: get_transform_object: could not find match for var for={(.*?)}; bug

untilhamza:
## What is this PR for:
- Handles issue #1728.
- Instead of raising an error, we return an empty list and log the error in the logger using logger.debug.

### DIFF
--- a/pytube/cipher.py
+++ b/pytube/cipher.py
@@ -218,6 +218,7 @@ def get_transform_object(js: str, var: str) -> List[str]:
     regex = re.compile(pattern, flags=re.DOTALL)
     transform_match = regex.search(js)
     if not transform_match:
+        logger.debug(f"No match found for pattern: {pattern}")
         return []  # Return an empty list if no match is found
 
     return transform_match.group(1).replace("\n", " ").split(", ")
--- a/pytube/cipher.py
+++ b/pytube/cipher.py
@@ -30,14 +30,12 @@ class Cipher:
         var_regex = re.compile(r"^\w+\W")
         var_match = var_regex.search(self.transform_plan[0])
         if not var_match:
-            raise RegexMatchError(
-                caller="__init__", pattern=var_regex.pattern
-            )
+            raise RegexMatchError(caller="__init__", pattern=var_regex.pattern)
         var = var_match.group(0)[:-1]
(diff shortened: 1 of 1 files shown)


BuildWithData:
this is great, thanks for sharing it !!! Question: have you managed to run the test suite ? tried but without success https://youtu.be/N0_ctnsKlmw

MarvinSchenkel:
Thanks for sharing! I just tried it, but using this PR as a dependency now results in the following error:

`pytube.exceptions.RegexMatchError: parse_function: could not find match for js_func_patterns`

Any ideas?

BuildWithData:
hey @MarvinSchenkel I patched pytube with this pull and it works fine for me but I have some changes already in place, you can check all the files I have patched [here](https://github.com/BuildWithData/Code-From-Videos/tree/master/pytube/patch). If you are having troubles patching, I made a [tutorial](https://youtu.be/Or7aNnrLF6s ) on how to do it

MarvinSchenkel:
> 

I use `pytube` as a dependency, so I do not manually patch but just reference this PR as a dependency instead of the `main` branch.

BuildWithData:
ah I see but I guess its not working bcs you need multiple pull requests, anyway patching is really easy, give it a go but if you wanna still go as you do it, [here](https://github.com/BuildWithData/Code-From-Videos/blob/master/pytube/patch/ReadMe.txt) it is the list of pull requests I used so far

MarvinSchenkel:
> ah I see but I guess its not working bcs you need multiple pull requests, anyway patching is really easy, give it a go but if you wanna still go as you do it, [here](https://github.com/BuildWithData/Code-From-Videos/blob/master/pytube/patch/ReadMe.txt) it is the list of pull requests I used so far

That actually makes a lot of sense. Do you happen to have a fork of `pytube` available somewhere with the latest patches applied?

glubsy (changes requested):
This fixes nothing at all, except *trying* to silently ignore the error for potentially more errors to crop up down the line. 

The transform_plan property is still expected to have at least one item in the returned list if there was no error during regex matching (see `Cipher` constructor).

PR #1734, opened 2023-07-19
remove corrupt function_patterns

mhi-iranmanesh:


glubsy (changes requested):
Please explain why this pattern is deemed "corrupt". 
Adding unit tests would help in that regard.

mhi-iranmanesh:
I was getting errors in some videos. After that I deleted the pattern and was able to get the videos.

PR #1738, opened 2023-07-22
Added get_download_url Function

abhaymakes:
Added `get_download_url` function in `streams.py` that returns the URL of the Video you're downloading.


santomon:
why not just use self.url directly?

abhaymakes:
Well yeah, but a function would be a better way to access the variable rather than directly accessing it.

Issue #1751, opened 2023-08-09
[Feature] Stopping a download in the middle


dani-rifkind:
Sometimes you want to stop the download in the middle(ie if you are making a interface of sorts)

Add a function that will change the value of a bool that will stop the download, by adding a check to see if the bool is positive or negative in between downloading each chunk.

PR #1752, opened 2023-08-09
Enable stopping the download in the middle.

dani-rifkind:
(https://github.com/pytube/pytube/issues/1751)
By calling the stop_download function in a Streams object you make the download function exit at the next possible chance(ie between chunks). Take note that the partial file will still exist.

### DIFF
--- a/pytube/cipher.py
+++ b/pytube/cipher.py
@@ -27,7 +27,8 @@ logger = logging.getLogger(__name__)
 class Cipher:
     def __init__(self, js: str):
         self.transform_plan: List[str] = get_transform_plan(js)
-        var_regex = re.compile(r"^\w+\W")
+        # var_regex = re.compile(r"^\w+\W")
+        var_regex = re.compile(r"^\$*\w+\W")
         var_match = var_regex.search(self.transform_plan[0])
         if not var_match:
             raise RegexMatchError(
--- a/pytube/cipher.py
+++ b/pytube/cipher.py
@@ -88,20 +88,20 @@ class Cipher:
         signature = list(ciphered_signature)
 
         for js_func in self.transform_plan:
-            name, argument = self.parse_function(js_func)  # type: ignore
-            signature = self.transform_map[name](signature, argument)
-            logger.debug(
-                "applied transform function\n"
-                "output: %s\n"
-                "js_function: %s\n"
-                "argument: %d\n"
-                "function: %s",
-                "".join(signature),
-                name,
-                argument,
-                self.transform_map[name],
-            )
-
+            if js_func:
+                name, argument = self.parse_function(js_func)  # type: ignore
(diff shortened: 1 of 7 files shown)


pytube/cipher.py
```diff
@@ -27,7 +27,8 @@
 class Cipher:
     def __init__(self, js: str):
         self.transform_plan: List[str] = get_transform_plan(js)
-        var_regex = re.compile(r"^\w+\W")
+        # var_regex = re.compile(r"^\w+\W")
+        var_regex = re.compile(r"^\$*\w+\W")
```
glubsy, line 31: Unrelated to this PR

pytube/streams.py
```diff
@@ -37,6 +37,9 @@ def __init__(
             Dictionary of data shared across all instances of
             :class:`Stream <Stream>`.
         """
+        # An extra attribute I added:
+        self.stop_signal = False
+        #################
```
glubsy: Remove comments

pytube/streams.py
```diff
@@ -316,6 +319,10 @@ def download(
                     timeout=timeout,
                     max_retries=max_retries
                 ):
+                    # An extra check I added:
+                    if self.stop_signal:
+                        break
+                    ##############
```
glubsy: remove comments

pytube/streams.py
```diff
@@ -334,8 +341,10 @@ def download(
                     # send to the on_progress callback.
                     self.on_progress(chunk, fh, bytes_remaining)
         self.on_complete(file_path)
+        self.stop_signal = False
```
glubsy, line 340: Why is this set again here?

pytube/streams.py
```diff
@@ -334,8 +341,10 @@ def download(
                     # send to the on_progress callback.
                     self.on_progress(chunk, fh, bytes_remaining)
         self.on_complete(file_path)
+        self.stop_signal = False        
         return file_path
-
+    def stop_download(self):
+        self.stop_signal = True
```
glubsy, line 344: Formatting (new lines) and what calls this? Is there signal handling?

glubsy (changes requested):

pytube/cipher.py
dani-rifkind: It fixes a different issue: https://github.com/pytube/pytube/issues/1768

pytube/streams.py
dani-rifkind: In case you want to download again using the same Stream object. Otherwise it wouldn't run

pytube/streams.py
dani-rifkind: The call is made using a Stream object by the user, What do you mean by signal handling?

Zebra-0:
tankyou, your code on streams really helped me out. works very well

PR #1758, opened 2023-08-12
Update cipher.py

BR7RKR:
fixed __init__: could not find match for ^\w+\W


kourdroid:
good the same problem was happen to me

Ripax:
this update already done in #1752

PR #1765, opened 2023-08-14
Fixed the regex error in cipher.py

SmileyTheSmile:


theaddonn:
works thx!!

perfecto25:
awesome was just looking at same issue, your fix is working

PR #1859, opened 2024-01-08
[CODE REFACTO] Create a status enum for code clarity

VassiaB-padoa:
Create a status enum for code clarity


glubsy (commented):
This does not add much in term of clarity. 
If the enum values were used in more than just this one place, then maybe it would make sense to have this enum.

PR #1863, opened 2024-01-13
Fixed the Description bug

Prithvi824:
Since the video info was not returning anything related to description, I created another http get request to fetch the description using regex.

### DIFF
--- a/pytube/__main__.py
+++ b/pytube/__main__.py
@@ -9,6 +9,7 @@ smaller peripheral modules and functions.
 import logging
 from typing import Any, Callable, Dict, List, Optional
 
+import re
 import pytube
 import pytube.exceptions as exceptions
 from pytube import extract, request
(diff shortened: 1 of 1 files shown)


pytube/__main__.py
```diff
@@ -362,8 +363,24 @@ def description(self) -> str:
...
+        end = re.compile(r'iscrawl\w+', re.IGNORECASE)
+
+        # Finds the pattern in the page
+        match1 = start.findall(page_detail)
+        match2 = end.findall(page_detail)
+
+        if match1 and match2:
+            # Retrives the index and trims not required commas
+            ind1 = page_detail.index(''.join(match1)) + 19
+            ind2 = page_detail.index(''.join(match2)) - 3
```
glubsy: What are these magic numbers?

glubsy (commented):
Why get rid of getting data from `vid_info` altogether? It should at least be used as a fallback.

Also, not really fond of doing requests inside a property getter. It should be a separate function/method, and I also think it should be called as a fallback if nothing is found in `vid_info`.

pytube/__main__.py
Prithvi824: Initially, I considered creating another request and then retrieving the description from the HTML. However, I discovered that "self.watch_html" also contains the description between "Shortdescription" and "iscrawlable." Therefore, I utilized a regex to find their indexes and return anything in between them, which constitutes the description. The addition and subtraction of 3 is for removing the colon and double quotes to get the text only.

The latest commit "Removed Internal request for Description - 2" consist of all the changes i mentioned

Prithvi824:
Using vid_info wouldn't be useful as it doesn't contain relevant information regarding the description.

Prithvi824:
Any Update??

glubsy (approved):
Looks good to me

Prithvi824:
Waiting for it to get merged then.

PR #1860, opened 2024-01-10
Update captions.py

justSam13:
This PR fixes the error mentioned in #1085

PR #1880, opened 2024-02-14
Update captions.py

OceanRamen:
Fixed errors in xml_caption_to_srt()


AdventurousSam:
Check out #1860 , I also added some code so that this function works for auto-generated captions also! :)

PR #1884, opened 2024-02-25
changed requests

nheffelman:


nheffelman:
allows me to use in kivy/android environment

PR #1906, opened 2024-04-03
Updated YouTube XML format

swigls:
YouTube XML format seems to be changed recently,
thus updated that changes in format to the code.
(If this update is only observed by me, please just ignore and close this PR)

The "changes" I updated:
* Timestamped elements location: child of `root` -> child of `body`
* start time keyword: `start` -> `t`
* duration keyword: `dur` -> `d`

### DIFF
--- a/pytube/captions.py
+++ b/pytube/captions.py
@@ -83,14 +83,15 @@ class Caption:
         """
         segments = []
         root = ElementTree.fromstring(xml_captions)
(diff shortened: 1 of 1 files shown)


nicholaslocascio (approved):

SK1Y101:
Bump: This change would be super helpful!
Seems to address #1085, which is the same I've encountered.

Issue #1911, opened 2024-04-15
[BUG] can't download this video


a26364177:
**To Reproduce**
pytube https://www.youtube.com/watch\?v\=J5Z7tIq7bco\&ab_channel\=SukimaSwitchVEVO
pytube https://www.youtube.com/watch?v=J5Z7tIq7bco

**Output**
```
Loading video...
Kanade.mp4 | 5 MB
Traceback (most recent call last):
  File "/Users/david/path/to/venv/bin/pytube", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/david/path/to/venv/lib/python3.12/site-packages/pytube/cli.py", line 53, in main
    _perform_args_on_youtube(youtube, args)
  File "/Users/david/path/to/venv/lib/python3.12/site-packages/pytube/cli.py", line 60, in _perform_args_on_youtube
    download_highest_resolution_progressive(
  File "/Users/david/path/to/venv/lib/python3.12/site-packages/pytube/cli.py", line 479, in download_highest_resolution_progressive
    _download(stream, target=target)
  File "/Users/david/path/to/venv/lib/python3.12/site-packages/pytube/cli.py", line 263, in _download
    stream.download(output_path=target, filename=filename)
  File "/Users/david/path/to/venv/lib/python3.12/site-packages/pytube/streams.py", line 314, in download
    for chunk in request.stream(
  File "/Users/david/path/to/venv/lib/python3.12/site-packages/pytube/request.py", line 189, in stream
    chunk = response.read()
            ^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/http/client.py", line 495, in read
    s = self._safe_read(self.length)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/http/client.py", line 642, in _safe_read
    raise IncompleteRead(data, amt-len(data))
http.client.IncompleteRead: IncompleteRead(120972 bytes read, 5782963 more expected)
**System information**
Please provide the following information:
- Python version (run `python --version`)
- Pytube version (run `print(pytube.__version__)` in python)
- Command used to install pytube

josuesalomao:
Hello! Can you send the code here?

SarthakNikhal:
[...]

PR #1913, opened 2024-04-20
Fix   can't download this video 1911

SarthakNikhal:
Fix for #1911
- This commit moves a tiny part of the stream function so that the part gets involved in exception handling of the http responses. This handles the IncompleteRead Exception which was used only for the response and not the data chunk extracted from the response. 
- This change will also bring some encapsulation of the response and chunk extracting process.

### DIFF
--- a/pytube/request.py
+++ b/pytube/request.py
@@ -159,6 +159,11 @@ def stream(
                     method="GET",
                     timeout=timeout
                 )
+                chunk = response.read()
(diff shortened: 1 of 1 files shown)


SarthakNikhal:
@RONNCC Let me know if any further work is required following this. I have tested my changes by downloading different types of videos. If you see a flaw. Please me ping me. Cheers!

PR #1922, opened 2024-05-07
Adapts _extract_videos function to new Youtube interface.

alrodsa:


alrodsa:
With this change, it will be possible to extract videos from a channel.

PR #1927, opened 2024-05-09
BUGFIX: Fix regex issue 

mahdjourOussama:
fix the regex issue in the cipher module that causes the package to stop working properly [https://github.com/pytube/pytube/issues/1918](issue related to it )


JayK31:
Assuming this needs to be done in tandem with the updates to the `get_throttling_function_name` from [here](https://github.com/pytube/pytube/pull/1691/files)?

PR #1961, opened 2024-07-12
Update setup.py

zinzied:
To discuss, it shows me error on line 18 so I imported  Ver. Function from setuptools lib.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 import codecs
 import os
 
-from setuptools import setup
+from setuptools import __version__, setup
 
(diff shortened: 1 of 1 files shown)


zstyblik:
I don't know which error and where you were, are, getting, however `__version__` in `version=__version__,  # noqa: F821` should be set by(from):

```
with open(os.path.join(here, "pytube", "version.py")) as fp:
    exec(fp.read())
```

If I'm reading it correctly.

`from setuptools import __version__, ...` imports and will set pytube version to version of `setuptools`.  Well, probably not because of the code above, but then this change doesn't make sense.

PR #1971, opened 2024-07-16
fix: 400 bad request

ahmedHanzala:


pikho:
This patch works, thanks!

PR #2153, opened 2025-07-21
fix json parsing bug for the continuation token variable

ConnorLAdams:
added a nested try accept for retrieving the continuation token through parsing. This keeps the functionality of the program mostly the same with one final check for the token.


HostServer001:
are you not getting a 400 bad request error ?

ConnorLAdams:
I am not

